PR #220, opened 2015-01-24
Extract binary filepath detection into separate modules

sindresorhus:
I've extracted the binary filepath detection into separate modules so other projects can benefit from the list of binary file extensions too.

https://github.com/sindresorhus/is-binary-path
https://github.com/sindresorhus/binary-extensions

I've added you both to the above modules on GitHub and npm.


test.js
```diff
@@ -979,11 +979,4 @@ describe('is-binary', function() {
   it('should be a function', function() {
     isBinary.should.be.a('function');
   });
-  it('should correctly determine binary files', function() {
```
sindresorhus: These passed before I removed them. No point in having them here as they're in `is-binary-path`.

lib/index.js
```diff
@@ -8,7 +8,7 @@ var globparent = require('glob-parent');
 
 var NodeFsHandler = require('./nodefs-handler');
 var FsEventsHandler = require('./fsevents-handler');
-exports.isBinaryPath = require('./is-binary');
+exports.isBinaryPath = require('is-binary-path');
```
sindresorhus: Any reason you expose this? I couldn't find it in the docs.

es128: Legacy. It's just been there forever and I've avoided any breaking API changes since I took over. But AFAICT nobody uses it, so now would be the time to remove it from here.

es128:
Thanks! LGTM

silverwind:
Just a friendly FYI: I like using [isBinaryFile](https://github.com/gjtorikian/isBinaryFile) for binary detection which doesn't rely on (hard to maintain) file extensions :)

es128:
Yeah but proper detection with `fs` operations is too heavy for this use-case.

silverwind:
True, it hurts performance.

PR #249, opened 2015-03-04
File size changed when using polling mode

edi9999:


lib/nodefs-handler.js
```diff
@@ -168,7 +168,7 @@ function setFsWatchFileListener(path, fullPath, options, handlers) {
           rawEmitter('change', fullPath, {curr: curr, prev: prev});
         });
         var currmtime = curr.mtime.getTime();
-        if (currmtime > prev.mtime.getTime() || currmtime === 0) {
+        if (curr.size != prev.size || currmtime > prev.mtime.getTime() || currmtime === 0) {
```
es128: `!==`

I know it shouldn't matter, but just for code style consistency

es128:
I don't see a problem conceptually with merging this. Not aware of any case where this could cause undesired behavior.

Would ask for a corresponding test, but I don't see a straightforward way to write a test for this. Do you?

edi9999:
I have been able to reproduce the issue by using bash piping on windows. Eg here is my code:

https://github.com/edi9999/test-chokidar

when I run out.js with `node out.js > watched/1.txt`, and I watch the `watched/` directory in `index.js` with polling, the mtime doesnt change, so the watch doesnt get triggered. 

However, I'm not sure that it is possible to add a test for this (especially because this seems to be windows only)

edi9999:
My mistake, the code doesn't reproduce the issue (the watches are triggered every 5 seconds in that case)

PR #252, opened 2015-03-05
[Work in Progress] Rewrite in ES6

steelbrain:
### The Problem

How to manage the es6 code in the repository

### The Options

 1. Update the source on each commit
 2. Compile The Source upon install.

I personally prefer the 1st option, It's easy to create a bash file in the project root that you can execute during development, It'll automatically setup Git Commit hooks that will compile the source files using `babel` and will add them to the Git Repo and they'll be committed along with Source Changes.

and the Source ES6 files can be npm ignored to keep the package clean.

@es128 which option should we choose?


es128:
Thanks for putting this together. However, I warned you before you started this that neither of those options is satisfactory. Checking in built files with every change is too inconvenient to be worth the trouble IMO, and requiring client-side compiling when installing from npm is totally unacceptable.

A `prepublish` hook could be ok, and probably instructions for users installing from git rather than a `postinstall`. A `postinstall` like we recently worked on at https://github.com/brunch/brunch/pull/926 might be considered (basically a noop when installed from the npm registry), but I'd be nervous about that as well considering chokidar's wide user base. 

Even listing babel in `devDependencies` doesn't sound so great as most users are installing chokidar in their dev environments and that's a much heavier dependency than anything we have now. I think keeping chokidar lightweight to install has some value. Alternately, requiring chokidar developers to install `babel` globally and working git/npm hook scripting around that might be alright.

Finally, if this is to be considered, it would have to be in pure ES6 as a pathway toward the source eventually running natively when targeted js engines fully support it. I would not consider using extended JSX features here at this point.

steelbrain:
> Finally, if this is to be considered, it would have to be in pure ES6 as a pathway toward the source eventually running natively when targeted js engines fully support it. I would not consider using extended JSX features here at this point.

Agreed. 'cause io.js supports classes natively and If we write this in ES6 instead of JSX, there's a possibility that we won't have to transpile it anymore in the near future.

> Even listing babel in devDependencies doesn't sound so great as most users are installing chokidar in their dev environments and that's a much heavier dependency than anything we have now. I think keeping chokidar lightweight to install has some value. Alternately, requiring chokidar developers to install babel globally and working git/npm hook scripting around that might be alright.

Agreed again.

> A prepublish hook could be ok, and probably instructions for users installing from git rather than a postinstall. A postinstall like we recently worked on at brunch/brunch#926 might be considered (basically a noop when installed from the npm registry), but I'd be nervous about that as well considering chokidar's wide user base.

Great idea. I'll do this in my other ES6 using projects as well.

es128:
We should also look at any other widely-used npm-published ES6-authored libraries we can find to see how they set up their tooling. Babel itself would be a good starting point, but I'd be interested in others that have more similarity to chokidar in terms of having a lightweight code base and dep graph.

steelbrain:
I am abandoning this PR. 'cause I no longer use chokidar. I've created my own [fs-watcher][fs] in less than 250 lines of code with about the same features and zero deps.
[fs]:https://github.com/steelbrain/node-fswatcher

PR #265, opened 2015-03-25
Check existence of emptied dirs

### LINKED ISSUE #234: Windows: unlinkDir sometimes doesn't fire
I still have to explore if this is reproducible, but I just deleted a directory with a single child file using

````js
rimraf(directory, {maxBusyTries: 10});
````

and while I got the `unlink` event for the file, no `unlinkDir` was fired for the parent directory. It does seem to happen pretty rarely, I encountered it in maybe one of hundred deletions.

### PR DESCRIPTION
**Pull request by es128**
Resolves #234

Couldn't write an effective test for this. Will merge if confirmed by issue reporter that it works.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -284,10 +284,18 @@ FSWatcher.prototype._getWatchHelpers = function(path, depth) {
 // Returns the directory's tracking object
 FSWatcher.prototype._getWatchedDir = function(directory) {
   var dir = sysPath.resolve(directory);
+  var watcherRemove = this._remove.bind(this);
   if (!(dir in this._watched)) this._watched[dir] = {
     _items: Object.create(null),
     add: function(item) {this._items[item] = true;},
-    remove: function(item) {delete this._items[item];},
+    remove: function(item) {
+      delete this._items[item];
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: lib/index.js [hunk 1]
```diff
@@ -284,10 +284,18 @@ FSWatcher.prototype._getWatchHelpers = function(path, depth) {
...
   if (!(dir in this._watched)) this._watched[dir] = {
     _items: Object.create(null),
     add: function(item) {this._items[item] = true;},
-    remove: function(item) {delete this._items[item];},
+    remove: function(item) {
+      delete this._items[item];
+      if (!this.children().length) {
+        fs.readdir(dir, function(err) {
+          if (err) watcherRemove(sysPath.dirname(dir), sysPath.basename(dir));
+        }.bind(this));
```
paulmillr: why two `bind`s

es128: oops that one isn't needed anymore... in an earlier iteration I hadn't aliased `this._remove`. Good catch

es128: a99d0057a7527f972647fdff22a39f9aaf15b0ab

PR #277, opened 2015-04-10
Update README.md - Add webpack to list of 'used by' projects

shakyShane:
You're doing yourself a disservice by not listing one of the most popular buzz-wordy tools out there at the moment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@ It is used in
 [socketstream](http://www.socketstream.org),
 [derby](http://derbyjs.com/),
 [watchify](https://github.com/substack/watchify),
+[webpack](https://github.com/webpack/webpack),
 and [many others](https://www.npmjs.org/browse/depended/chokidar/).
 It has proven itself in production environments.
 


shakyShane:
dep tree [webpack](https://github.com/webpack/webpack) -> [watchpack](https://github.com/webpack/watchpack) -> chokidar

es128:
Good point, thanks. 

I wonder if we should list browserify instead of watchify. Maybe change the text while the link still goes to the watchify repo, and then the same sort of thing with webpack/watchpack. 

Thoughts?

shakyShane:
If you consider that most users will have heard of `browserify` and `webpack` - but perhaps not their sub-deps or sister projects, I would suggest it would be clear to just list the main project names for max exposure.

es128:
And have the links point to where?

shakyShane:
The main project pages for each... ?

es128:
Seems slightly confusing since chokidar isn't a dep there. Maybe I'm overthinking it. 

But still, I'm kinda leaning toward the text listing the main project: webpack/browserify/gulp(eventually), while the link points to the repo that actually uses chokidar: watchify/watchpack/glob-watcher

shakyShane:
^ seems like the perfect compromise :+1:

es128:
Well I kind of changed my mind since all these projects have nice websites I decided to use those rather than repo links (other than PM2 since their website is about the commercial service rather than the open source project). Should be clear enough - anyone motivated to understand *how* these projects use chokidar should be able to figure it out.

PR #284, opened 2015-04-18
Add command line interface

kimmobrunfeldt:
Add cli too named `watch`. Argument parsing done with [yargs](https://github.com/bcoe/yargs). Adds a few dependencies.
```
➜  watch --help
Usage: watch <command> <pattern> [options]

<command>:
Command to be executed when a change is detected.
Needs to be surrounded with quotes when command contains spaces

<pattern>:
Glob pattern to specify files to be watched.
Needs to be surrounded with quotes to prevent shell globbing.
Guide to globs: https://github.com/isaacs/node-glob#glob-primer

Options:
  -d, --debounce          Debounce timeout in ms for executing command
                                                                  [default: 400]
  -s, --follow-symlinks   When not set, only the symlinks themselves will be
                          watched for changes instead of following the link
                          references and bubbling events through the links path
                                                     [boolean]  [default: false]
  -i, --ignore            Pattern for files which should be ignored. Needs to
                          be surrounded with quotes to prevent shell globbing.
                          The whole relative or absolute path is tested, not
                          just filename
  -p, --polling           Whether to use fs.watchFile(backed by polling)
                          instead of fs.watch. This might lead to high CPU
                          utilization. It is typically necessary to set this to
                          true to successfully watch files over a network, and
                          it may be necessary to successfully watch files in
                          other non-standard situations
                                                     [boolean]  [default: false]
  --poll-interval         Interval of file system polling. Effective when
                          --polling is set                        [default: 100]
  --poll-interval-binary  Interval of file system polling for binary files.
                          Effective when --polling is set         [default: 300]
  -h, --help              Show help
  -v, --version           Show version number

Examples:
  watch "npm run build-js" "**/*.js"    build when any .js file changes
```

### DIFF
--- a/cli/watch.js
+++ b/cli/watch.js
@@ -1,0 +1,161 @@
+#!/usr/bin/env node
+
+var childProcess = require('child_process');
+var Promise = require('bluebird');
+var _ = require('lodash');
+var minimatch = require('minimatch')
+var chokidar = require('../index');
+
+var defaultOpts = {
+    debounce: 400,
+    followSymlinks: false,
+    ignore: null,
+    polling: false,
+    pollInterval: 100,
+    pollIntervalBinary: 300
+};
+
+var argv = require('yargs')
+    .usage(
+        'Usage: $0 <command> <pattern> [options]\n\n' +
+        '<command>:\n' +
+        'Command to be executed when a change is detected.\n' +
+        'Needs to be surrounded with quotes when command contains spaces\n\n'+
+        '<pattern>:\n' +
+        'Glob pattern to specify files to be watched.\n' +
+        'Needs to be surrounded with quotes to prevent shell globbing.\n' +
+        'Guide to globs: https://github.com/isaacs/node-glob#glob-primer'
+    )
+    .example('$0 "npm run build-js" "**/*.js"', 'build when any .js file changes')
+    .demand(2)
+    .option('d', {
+        alias: 'debounce',
+        default: defaultOpts.debounce,
+        describe: 'Debounce timeout in ms for executing command',
+        type: 'number'
+    })
+    .option('s', {
+        alias: 'follow-symlinks',
+        default: defaultOpts.followSymlinks,
+        describe: 'When not set, only the symlinks themselves will be watched ' +
+                  'for changes instead of following the link references and ' +
+                  'bubbling events through the links path',
+        type: 'boolean'
+    })
+    .option('i', {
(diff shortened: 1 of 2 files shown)


paulmillr:
`watch` sounds too generic, `chokidar` would be better here. You can always do `alias watch=chokidar`, but you cannot do otherwise if you have bin file conflicts.

But anyway, I think the binary is out of scope for chokidar, an NPM interface. I would suggest to create your own NPM package that'll do the same thing. We can mention it in README.

es128:
Thanks for the contribution @kimmobrunfeldt, but I agree with @paulmillr that this belongs in a separate package.

Just as a side note, we try to minimize the dependencies in chokidar, so even if we wanted to merge this feature would have needed to be changed to drop the convenience features of the lodash and bluebird utility libs. Minimatch was added without being used, and chokidar uses anymatch for glob matching anyway. I would have been ok with keeping yargs or an alternate argument parsing lib to support this feature though.

Also `watch` as a node CLI command already exists: https://github.com/mikeal/watch

It's worth noting there are some existing CLI wrappers for chokidar on npm already

https://www.npmjs.com/package/onchange
https://www.npmjs.com/package/node-fwatch
https://www.npmjs.com/package/dirwatch
https://www.npmjs.com/package/chokidar-cmd
https://www.npmjs.com/package/chok

@paulmillr if we're going to mention this in the readme we should pick one favorite. Probably the first one. What do you think?

paulmillr:
Let's use the first one.

kimmobrunfeldt:
I created a new cli which supports quite a lot of options: https://github.com/kimmobrunfeldt/chokidar-cli

kimmobrunfeldt:
It is not fully ready yet but will be in the following few days. I'm going to use it my self in my other projects so I'll keep maintaining it.

kimmobrunfeldt:
@paulmillr @es128 is now "ready". https://www.npmjs.com/package/chokidar-cli

I renamed the binary to `chokidar`.

kimmobrunfeldt:
I added support for piping/change streaming and moved "run command on change" under optional `-c, --command` flag. Now it's more general purpose tool.

es128:
@kimmobrunfeldt sounds good. I'm warming up to the idea of recommending your tool in the README. 

I think it would be cool if you could add @paulmillr and I as collabs on the repo & npm.

kimmobrunfeldt:
@es128 sure, done.

PR #291, opened 2015-05-04
Cleanup watcher's state after unwatching a path (fixes #289)

### LINKED ISSUE #289: No events for paths that were watched > unwatched > added again
I'm facing an issue of having some paths that are watched, I unwatch them and then I add them again for watching.

**OS**: Linux 3.19.3 on Debian stretch/sid
**filesystem**: ext4
**nodejs**: v0.10.29

He're is a patch adding a failing test case  in `test.js` (at least in my env):
```diff
diff --git a/test.js b/test.js
index 255359b..dd5437e 100644
--- a/test.js
+++ b/test.js
@@ -1234,6 +1234,24 @@ function runTests(options) {
           }));
       })();
     });
+    it('should watch paths the were unwatched and added again', function(done) {
+      var spy = sinon.spy();
+      var watchPaths = [getFixturePath('add.txt'), getFixturePath('change.txt')];
+      watcher = chokidar.watch(watchPaths, options)
+        .on('all', spy)
+        .on('ready', d(function() {
+          watcher.unwatch(getFixturePath('change.txt'));
+          watcher.add(getFixturePath('change.txt'));
+          fs.writeFileSync(getFixturePath('add.txt'), 'c');
+          fs.writeFileSync(getFixturePath('change.txt'), 'c');
+          waitFor([spy, 2], function() {
+            spy.should.have.been.calledWith('add', getFixturePath('add.txt'));
+            spy.should.have.been.calledWith('change', getFixturePath('change.txt'));
+            spy.should.have.been.calledTwice;
+            done();
+          });
+        }));
+    });
   });
   describe('close', function() {
     beforeEach(clean);
```

mehiel:
This should address the case with #289. In my working example it really worked after this change.

@es128, can you please check what's wrong with the test? It seems correct but it probably isn't, it doesn't pass in my env. As I said the change in `index.js` seems to indeed solve the issue, I double checked it with my production code. I also ran it under strace and it's now calling the `inotify_add_watcher` syscall again after remove, which was not the case before the fix. But the test is still failing :(

### DIFF
--- a/index.js
+++ b/index.js
@@ -430,7 +430,7 @@ FSWatcher.prototype.unwatch = function(paths) {
     if (this._closers[path]) {
       this._closers[path]();
       delete this._closers[path];
-      delete this._getWatchedDir(sysPath.dirname(path)).remove(sysPath.basename(path));
+      this._getWatchedDir(sysPath.dirname(path)).remove(sysPath.basename(path));
     } else {
       this._ignoredPaths[path] = true;
       if (path in this._watched) this._ignoredPaths[path + '/**/*'] = true;
--- a/index.js
+++ b/index.js
@@ -429,6 +429,8 @@ FSWatcher.prototype.unwatch = function(paths) {
   paths.forEach(function(path) {
     if (this._closers[path]) {
       this._closers[path]();
+      delete this._closers[path];
+      delete this._getWatchedDir(sysPath.dirname(path)).remove(sysPath.basename(path));
     } else {
       this._ignoredPaths[path] = true;
       if (path in this._watched) this._ignoredPaths[path + '/**/*'] = true;
--- a/test.js
+++ b/test.js
@@ -1234,6 +1234,23 @@ function runTests(options) {
           }));
       })();
     });
+    it('should watch paths the were unwatched and added again', function(done) {
+      var spy = sinon.spy();
+      var watchPaths = [getFixturePath('change.txt')];
+      watcher = chokidar.watch(watchPaths, options)
+        .on('all', spy)
+        .on('ready', d(function() {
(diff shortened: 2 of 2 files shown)


mehiel:
Interesting, it does pass in Travis but _'should ignore unwatched paths that are a subset of watched paths'_ has problems there.

index.js
```diff
@@ -429,6 +429,8 @@ FSWatcher.prototype.unwatch = function(paths) {
   paths.forEach(function(path) {
     if (this._closers[path]) {
       this._closers[path]();
+      delete this._closers[path];
+      delete this._getWatchedDir(sysPath.dirname(path)).remove(sysPath.basename(path));
```
es128: Why is there a `delete` on this line? I think perhaps just `this._getWatchedDir(sysPath.dirname(path)).remove(sysPath.basename(path));` instead of both these lines would solve the issue without side effects.

(I suspect deleting the closer is a neutral change, not harmful, but also not necessary)

Haven't yet tried it out, but want to carve out some time today to do so.

mehiel: Consider this `delete` an **oops**. I've missed when I merged some local branches and had a conflict. Sorry about that.

mehiel: Yes _closers do not harm but I though it would be better to clean that part too.

es128:
Turns out that wasn't the issue apparently. Something about that other test seems to be depending on the path's key still being there.

mehiel:
Indeed. I'll try to find some time later in the day to see if I can find what's wrong with that test. Thanks!

es128:
All tests are passing locally using this patch (rebased onto current master). Going to go ahead and merge and then see where we are with the CI test results

PR #311, opened 2015-07-13
Ignore files when using the 'cwd' option

JimRobs:
I ran into a minor issue when files are not ignored when using `cwd` option with `chokidar.watch`, neither in the `paths` parameter with paths starting with `!`, nor in the `ignored` property of the `options` parameter.

I fixed it jus taking care of the `cwd` option in the `_isIgnored` function and change taking care of paths starting with `!` in the `add` function when `cwd` is used.

I also tested this feature in the *cwd describe* in the **test.js**.

### DIFF
--- a/index.js
+++ b/index.js
@@ -206,17 +206,31 @@ FSWatcher.prototype._isIgnored = function(path, stats) {
     /\..*\.(sw[px])$|\~$|\.subl.*\.tmp/.test(path)
   ) return true;
 
-  // create the anymatch fn if it doesn't already exist
-  this._userIgnored = this._userIgnored || anymatch(this._globIgnored
-    .concat(this.options.ignored)
-    .concat(arrify(this.options.ignored)
-      .filter(function(path) {
-        return typeof path === 'string' && !isglob(path);
-      }).map(function(path) {
-        return path + '/**/*';
-      })
-    )
-  );
+  if(!this._userIgnored){
+    var cwd = this.options.cwd;
+    var ignored;
+    if(cwd){
+      if(!this.options.ignored){
+        ignored = [];
+      } else {
+        ignored = arrify(this.options.ignored).map(function(path){
+          return isAbsolute(path) ? path : sysPath.join(cwd, path);
+        });
+      }
+    } else {
+      ignored = this.options.ignored;
+    }
+    this._userIgnored = anymatch(this._globIgnored
+      .concat(ignored)
+      .concat(arrify(ignored)
+        .filter(function(path) {
+          return typeof path === 'string' && !isglob(path);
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -206,17 +206,31 @@ FSWatcher.prototype._isIgnored = function(path, stats) {
...
-    .concat(this.options.ignored)
-    .concat(arrify(this.options.ignored)
-      .filter(function(path) {
-        return typeof path === 'string' && !isglob(path);
-      }).map(function(path) {
-        return path + '/**/*';
-      })
-    )
-  );
+  if(!this._userIgnored){
```
es128: pls match code style on all the `if` statements (missing spaces)

index.js
```diff
@@ -206,17 +206,31 @@ FSWatcher.prototype._isIgnored = function(path, stats) {
...
+    if(cwd){
+      if(!this.options.ignored){
+        ignored = [];
+      } else {
+        ignored = arrify(this.options.ignored).map(function(path){
+          return isAbsolute(path) ? path : sysPath.join(cwd, path);
+        });
+      }
+    } else {
+      ignored = this.options.ignored;
```
es128: Couldn't this have just been done on line 211 and all the rest of these conditionals condensed to only `if (cwd && !this.options.ignored)`? Seems like it's doing the right thing overall, but the code could be refactored and simplified.

index.js
```diff
@@ -380,7 +394,13 @@ FSWatcher.prototype.add = function(paths, _origAdd, _internal) {
   paths = arrify(paths);
 
   if (cwd) paths = paths.map(function(path) {
-    return isAbsolute(path) ? path : sysPath.join(cwd, path);
+    if(isAbsolute(path)){
+      return path;
+    } else if(path[0] == '!'){
```
es128: `===`

index.js
```diff
@@ -380,7 +394,13 @@ FSWatcher.prototype.add = function(paths, _origAdd, _internal) {
...
+      return '!'+sysPath.join(cwd, path.substr(1));
```
es128: spaces around the `+` operator

index.js
es128: Why is this the second condition being checked instead of the first?

es128:
Thanks! Just a few code style nits, but generally looks good for merge+publish.

index.js
JimRobs: In all cases, the `cwd` and `path` should not be joined if the path is absolute, just like it was before.

index.js
JimRobs: I will also use `substring` instead of `substr` for consistency.

PR #332, opened 2015-08-10
Use fsevents 0.3.7 to avoid segmentation faults

pmq20:
cf.

* https://github.com/strongloop/fsevents/pull/74
* https://github.com/strongloop/fsevents/issues/52

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "sinon-chai": "^2.6.0"
   },
   "optionalDependencies": {
-    "fsevents": "^0.3.1"
+    "fsevents": "^0.3.7"
   },
   "dependencies": {
     "anymatch": "^1.1.0",


package.json
```diff
@@ -38,7 +38,7 @@
     "sinon-chai": "^2.6.0"
   },
   "optionalDependencies": {
-    "fsevents": "^0.3.1"
+    "fsevents": ">= 0.3.7 < 1"
```
fengmk2: `^0.3.7`

es128:
This is unnecessary. `^0.3.1` will already install the latest 0.3.x automatically, and the only thing existing projects need to do is `npm update`.

paulmillr:
Doesn't bring any disadvantages tho!

nickdesaulniers:
if you can publish a new version number, folks reinstalling watchify will get unblocked, see: https://github.com/substack/watchify/issues/257#issuecomment-132843293

es128:
@nickdesaulniers shouldn't matter. `npm update` (or at worst `rm -rf node_modules && npm i`) should be sufficient to pull in latest fsevents.

master branch is failing tests in CI currently, want to resolve before the next publish

nickdesaulniers:
yep, thanks!

es128:
> Doesn't bring any disadvantages tho!

Turns out it does. It confuses *lots* of users thinking they need to wait for us to publish rather than just updating deps and letting semver do its thing.

Issue #344, opened 2015-09-03
Folder with parenthesis in the name can't be watched


javiergarmon:
I've a folder called `"/Users/Javier/Backup (2015)"`. I can't start a watcher to this folder because inside `FSWatcher._getWatchHelpers`, the library `globparent` transform that path to `"/Users/Javier"`.

Is this a bug? How can I watch folders with special chars like pathensis, brackets...?

Thank you!

es128:
Duplicate of #300.

Yes it's a bug, but easiest resolution for now is to just rename your dir.

Issue #336, opened 2015-08-14
chokidar doesn't notice changes under weird path


ariovistus:
In windows 7 I attempt to run the following script:

```
chokidar = require('chokidar')
fs = require('fs')

path = "APCS (3.0)2\\burrito.txt"
console.info(fs.lstatSync(path)) # path exists
watcher = chokidar.watch(path)
watcher.on('change', (path) -> console.info("path: #{path} changed"))

setTimeout(( -> 1), 50000)
```

chokidar never notices file changes to burrito.txt.

es128:
likely a dupe of #300

Issue #300, opened 2015-06-04
Issue with watching directory with name including glob characters


bpasero:
- have a folder "foo*" (extra star at the end)
- try to watch it with chokidar
  => it will not work because foo\* is treated as glob pattern

How is it possible to use a path "as is" without glob smartness?

es128:
There is no option to disable glob matching. 

For filenames that use glob characters it should work as that issue has been reported in the past (by you iirc) and fixed. 

It may be necessary to look into a similar fix for directory names. 

However, I'm not sure why the example you've provided wouldn't already work. `foo*` as a glob pattern should match `foo*` as a directory name. Unless the issue is that it's also matching the `foobar` directory and you don't want that.

bpasero:
Yes, this is an issue for folders I want to watch on that contain glob
sensitive characters.
Am 04.06.2015 2:51 nachm. schrieb "Elan Shanker" <notifications@github.com>:

es128:
Can you provide a more realistic example?

bpasero:
Wait, what? How can you say a valid folder name which chokidar cannot watch is not a realistic example :)?

es128:
I'm saying that given `foo*` as a path to watch chokidar *would* watch a directory named `foo*` since that matches the glob pattern. 

This wouldn't be the case for any glob-like name, so I'm interested in your actual example.

bpasero:
Hm, I was running an example on my Mac with latest chokidar watching a folder "foo*" that was actually on disk (~/Desktop/foo*) and I did not receive file events. I can try again later when I am back at my Mac to post a snippet here, but I thought the issue was more obvious.

bpasero:
@es128 Here are the steps:

- have a folder on the desktop called bar with a folder called foo*
- run following code
- drop a file into the folder foo*
- you dont see an output for the event

```javascript
var chokidar = require("chokidar");

var watcher = chokidar.watch('/Users/<name>/Desktop/bar/foo*');
watcher.on("all", function(event, path) {
	console.log(event, path);
});
```

paulmillr:
The question here is: the `foo*` folder name is an actual real-world case in your project(s) or it's just a scenario?

bpasero:
The folder in question was "js* - rewrite" by a colleague of mine using Visual Studio Code, complaining file events do not work. So yeah, this is real!

ArtemGovorov:
We have hit the same issue in wallaby.js. Here is what we do:
``` JavaScript
[...]

PR #347, opened 2015-09-06
Added option to prevent path transformation

javiergarmon:
It fixes some issues like:
* Issue with watching directory with name including glob characters https://github.com/paulmillr/chokidar/issues/300
* chokidar doesn't notice changes under weird path https://github.com/paulmillr/chokidar/issues/336
* Folder with parenthesis in the name can't be watched https://github.com/paulmillr/chokidar/issues/344

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,9 +142,9 @@ symlinks themselves will be watched for changes instead of following
 the link references and bubbling events through the link's path.
 * `cwd` (no default). The base directory from which watch `paths` are to be
 derived. Paths emitted with events will be relative to this.
-* `preventPathTransformation` (default: `false`). When `true`, only the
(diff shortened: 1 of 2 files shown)


paulmillr:
Whatsup with those tests?

javiergarmon:
Are you asking about Travis tests or are you asking to write some tests for this functionality?

es128:
The tests are failing because of a pre-existing issue in master with `unwatch`. Trying to find time to work on debugging it soon. 

But nonetheless :-1: on merging this change. I agree it will alleviate some pain, but this is the easy way out and I believe it can be solved without adding a new option. 

Even if we were to decide to incorporate this, yes it definitely would need new tests to cover the added feature.

PR #349, opened 2015-09-08
Use fsevents 0.3.8 to ensure support for Node v4.

gdi2290:


TrySound:
https://github.com/paulmillr/chokidar/pull/332#issuecomment-129461277

es128:
As @TrySound pointed out, the solution is to `npm update` regardless of whether chokidar bumps the version.

I'll go ahead and merge though, since (as @paulmillr has pointed out) it isn't going to hurt anything. But please do not expect a new release of chokidar just for the sake of incrementing this patch version.

Issue #273, opened 2015-04-01
Unwatch is not unwatching folders


rfossella:
I'm using chokidar on Windows 7.  Installed latest version today 1.3.11

I have a long-running node process that needs to **watch** and **unwatch**  different folders daily. I successfully watch a collection of folders with all proper events.  However, when I try to **UNWATCH** folders and then **WATCH** new folders,  the **old** folders (now, presumably UNWATCHED) still generate notifications.  The test I currently use is to UNWATCH/WATCH after 30 seconds.  When I use watch.close() I (obviously) receive NO notifications.  Can someone tell what I may be missing?
Thanks.

```
var chokidar = require('chokidar');
var folders = ['tmp','tmp2','tmp3'];        

var watcher = chokidar.watch(folders, {
  ignored: /[\/\\]\./, persistent: true
});

watcher
    .on('add', function(path, stats) 
        { 
            console.log({type: 'File',  action: 'create', path: path , stats:stats });
        })
    .on('addDir', function(path, stats) 
        { 
            console.log({type: 'Directory',  action: 'create', path: path , stats:stats });
        })
    .on('change', function(path, stats) 
        { 
            console.log({type: 'File',  action: 'change', path: path , stats:stats });
        })
    .on('unlink', function(path, stats) 
        { 
            console.log({type: 'File',  action: 'delete', path: path , stats:stats });
        })
    .on('unlinkDir', function(path, stats) 
        { 
            console.log({type: 'Directory',  action: 'delete', path: path , stats:stats });
        })
    .on('error', function(error) 
        { 
            console.log('Error occurred ' + error);
        })
    .on('ready', function() 
        { 
            console.log('Initial scan complete. Ready for changes.', "blue");
        })

 // Unwatch/Add watch after 30 seconds 
setTimeout(function(){
    console.log('About to watch/unwatch our paths');
    watcher.add(['tmp4','tmp5']);
    watcher.unwatch(['tmp','tmp2','tmp3']);

    //watcher.close()

},30000); 
```
Any activity on folders  __['tmp','tmp2','tmp3']__ is still tracked, which I do not want nor expect.

es128:
What is different about your situation compared with what's being done in the passing tests at https://github.com/paulmillr/chokidar/blob/master/test.js#L1171-1215 ? If you run the tests in your environment, do they pass?

rfossella:
[...]

Issue #258, opened 2015-03-17
unwatched file still trigger "all" event when modified


MoOx:
Here is my scenario:

I watch 2 files A & B, I add them & execute a function on "all" event. If I unwatch B & modify this B file, I still have the watcher triggering "all" event, is this on purpose ? Looks like a bug.

To reproduce here is a branch: https://github.com/cssnext/cssnext/tree/watch-imported

Here is the part that add/unwtach appropriate files 

https://github.com/cssnext/cssnext/blob/fbfaff5053180192e146cdde8d1986a39e0d3980/bin/cssnext.js#L107-L113

Here is the broken test https://github.com/cssnext/cssnext/blob/fbfaff5053180192e146cdde8d1986a39e0d3980/test/cli.js#L177-L189

I have also tried by hand. To reproduce:
1. clone the repo, checkout the branch `watch-imported`, then `npm install`
- run the following command `bin/cssnext.js --watch test/fixtures/cli.watch-import.css test/fixtures/cli.watch-import.actual.css`
- modify `test/fixtures/cli.watch-import.css` with `/**/ @import "cli.watch-import-import.css";`: you will see that the imported file has been added
- trigger a change in this imported file `test/fixtures/cli.watch-import-import.css` (_this is working as expected_)
- now modify `test/fixtures/cli.watch-import.css` with `/**/`: you will see that the file is _unwatch_ed
- trigger a change in this (now supposed to be unwatched) previously imported file `test/fixtures/cli.watch-import-import.css`: you got a change triggered by the watcher (_this is not what I expect_)

es128:
Thanks for reporting. Can you please provide a script or procedure to reproduce using only chokidar? The example.js script in the repo is usually a good starting point.

es128:
For reference, here are the passing tests for `unwatch`: https://github.com/paulmillr/chokidar/blob/master/test.js#L1109

Perhaps modifying one of them to act on individual files like in your scenario will expose the issue

PR #352, opened 2015-09-12
Add waitWriteFinish option

alexpusch:
The `add` event will fire when a file first appear on disk, before the entire file has been written. Furthermore, in some cases some `change` events will be emitted while the file is being written.
In some cases, especially when watching for large files there will be a need to wait for the write operation to finish before responding to the file creation.

Current event throttling is insufficient to handle this case. For a large enough file, or slow enough write, change event might be emitted for several seconds after the file have appeared.

Setting `waitWriteFinish` to `true` will poll a newly created file size, holding its `add` and `change` events until its size will not change for `writeFinishThreshold` amount of time.

**Note**
Some tests failed to pass for me. I've patched them in the first commit, but I'm not confident in the fix. Please advise.
Fix might be related to issue #273 or issue #258

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,6 +167,20 @@ subdirectories will be traversed.
   polling for binary files.
   ([see list of binary extensions](https://github.com/sindresorhus/binary-extensions/blob/master/binary-extensions.json))
 
+#### Waiting for write operation to finish
+* `waitWriteFinish` (default: `false`).
+The `add` event will fire when a file first appear on disk, before the entire
+file has been written. Furthermore, in some cases some `change` events will be emitted while the file is being written.
+In some cases, especially when watching for large files there will be a need to 
+wait for the write operation to finish before responding to the file creation.
+Setting `waitWriteFinish` to `true` will poll a newly created file size, holding 
+its `add` and `change` events until its size will not change for a configurable amount of time.
+* `writeFinishThreshold` (default: 2000).
+Amount of time in milliseconds for a file size to remain constant before emitting its 
+`add` event. Unfortunately this duration is heavily dependent on OS and local hardware.
+For accurate detection this parameter should be relatively high, making file watching much less
+responsive. Use with cation. 
+
 #### Errors
 * `ignorePermissionErrors` (default: `false`). Indicates whether to watch files
 that don't have read permissions if possible. If watching fails due to `EPERM`
--- a/README.md
+++ b/README.md
@@ -168,18 +168,34 @@ subdirectories will be traversed.
   ([see list of binary extensions](https://github.com/sindresorhus/binary-extensions/blob/master/binary-extensions.json))
 
 #### Waiting for write operation to finish
-* `waitWriteFinish` (default: `false`).
+* `awaitWriteFinish` (default: `false`).
 The `add` event will fire when a file first appear on disk, before the entire
 file has been written. Furthermore, in some cases some `change` events will be emitted while the file is being written.
 In some cases, especially when watching for large files there will be a need to 
 wait for the write operation to finish before responding to the file creation.
-Setting `waitWriteFinish` to `true` will poll a newly created file size, holding 
-its `add` and `change` events until its size will not change for a configurable amount of time.
-* `writeFinishThreshold` (default: 2000).
-Amount of time in milliseconds for a file size to remain constant before emitting its 
-`add` event. Unfortunately this duration is heavily dependent on OS and local hardware.
+Setting `awaitWriteFinish` to `true` will poll a newly created file size, holding 
+its `add` and `change` events until the size does not change for a configurable amount of time.
+The appropriate duration setting is heavily dependent on the OS and hardware.
 For accurate detection this parameter should be relatively high, making file watching much less
 responsive. Use with cation. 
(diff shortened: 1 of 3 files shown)


es128:
Thanks for the PR. Looks like a useful option, and I think I'm alright with the solution concept.

I'll need to take some time to review carefully and try the functionality out before merging. 

And we do need to get your new tests passing on Travis, which [is not consistently the case now](https://travis-ci.org/paulmillr/chokidar/builds/80001124). It might have to do with how the synthetic delays are laid out, which can be really tricky to get right. I'd suggest adding more delay wrappers until the tests pass consistently, and then start removing ones that seem extraneous a little at a time watching the results with each small change.

alexpusch:
I've now passed my tests. There are some unrelated tests that fail. The failed tests actually helped me find an edge case that I didn't handle. You can see it in this [commit](https://github.com/alexpusch/chokidar/commit/62f948181d92f9b60278f213207cbb1a4a59b359). I'll rebase this commit to the main one after you review the code.

es128:
@alexpusch please rebase. Would also appreciate if you could take a pass at matching code style better - notably spacing in your `if` blocks and semicolons.

alexpusch:
I went over the style issues and updated with 1.0.6. Rebased all fixes into current commit

index.js
```diff
@@ -368,6 +434,12 @@ FSWatcher.prototype._remove = function(directory, item) {
   var wasTracked = parent.has(item);
   parent.remove(item);
 
+  // If we wait for this file to be fully written, cancel the wait.
+  if(this.waitWriteFinish && this._pendingWrites[path]){
```
paulmillr: space

paulmillr: why `this.waitWriteFinish` instead of `options.waitWriteFinish`

alexpusch: Rebased fix

es128:
@alexpusch I haven't found the time quite yet to play with this locally, but hope to get to it shortly.

In the meantime, I'm wondering why 2000 is an appropriate default - can't it be any lower?

The latest commit also reintroduces some code style issues with a bunch of `if(`s missing a space.

README.md
```diff
@@ -167,6 +167,20 @@ subdirectories will be traversed.
...
   ([see list of binary extensions](https://github.com/sindresorhus/binary-extensions/blob/master/binary-extensions.json))
 
+#### Waiting for write operation to finish
+* `waitWriteFinish` (default: `false`).
+The `add` event will fire when a file first appear on disk, before the entire
+file has been written. Furthermore, in some cases some `change` events will be emitted while the file is being written.
+In some cases, especially when watching for large files there will be a need to 
+wait for the write operation to finish before responding to the file creation.
+Setting `waitWriteFinish` to `true` will poll a newly created file size, holding 
+its `add` and `change` events until its size will not change for a configurable amount of time.
```
es128: until the size does not change for a configurable amount of time

es128: The "Use with caution." sentence should be moved up here.

README.md
```diff
@@ -167,6 +167,20 @@ subdirectories will be traversed.
...
+* `writeFinishThreshold` (default: 2000).
+Amount of time in milliseconds for a file size to remain constant before emitting its 
+`add` event. Unfortunately this duration is heavily dependent on OS and local hardware.
```
es128: Don't use the word "unfortunately"

The appropriate duration setting is heavily dependent on the OS and hardware.

index.js
```diff
@@ -194,6 +215,51 @@ FSWatcher.prototype._throttle = function(action, path, timeout) {
...
+      if(this._pendingWrites[path] === undefined){
+        this._pendingWrites[path] = {
+          creationTime: now,
+          cancelWait: function(){
+            delete this._pendingWrites[path];
+            clearTimeout(timeoutHandler);
+            return callback();
+          }.bind(this)
+        }
+        return timeoutHandler = setTimeout(awaitWriteFinish.bind(this, curStat), this.options.interval);
```
es128: Conflating the use of the `interval` option here, which is intended (and documented) as applying to the `usePolling` option. This probably needs a dedicated property for `awaitWriteFinishInterval`.

es128: Maybe the options need to be laid out more like

```js
awaitWriteFinish: {
    enabled: true,
    pollInterval: 250,
    stabilityThreshold: 2000
}
```

alexpusch:
The number 2000 came of some local testing I've done. I've noticed that simple cp or mv commands tend to have 'pauses' of a second or two during the write process. It has a lot to do with the speed/number of disks you have, RAID setting, etc. I'll research it some more.

As for the option structure - Currently there is no other option that have its options encapsulated in a dedicated object. I can see how this approach tidying up the options object. Do you think it's worth going out of convention? 

Maybe the use of this option with the default values can be a simple
```js
awaitWriteFinish: true
```
And if the user wishes to change some parameters:
```js
awaitWriteFinish: {
   pollInterval: 250,
   stabilityThreshold: 2000
}
```

es128:
:+1: I'm good with that suggestion for the config

As for the default values, if those are the values that work best for you, I'm ok with sticking to that for now and seeing what happens when we get feedback from other users of this feature.

PR #362, opened 2015-09-30
Handle nested arrays

### LINKED ISSUE #360: Does not handle nested arrays of paths
In index.js line 241:

```javascript
path = path.replace(/^\.[\/\\]/, '');
```

This is throwing the following error: 

```
TypeError: path.replace is not a function
    at EventEmitter.FSWatcher._getWatchHelpers (/Users/Adam/Sites/clients/mka/node_modules/gulp/node_modules/chokidar/index.js:241:15)
    at EventEmitter.FsEventsHandler._addToFsEvents (/Users/Adam/Sites/clients/mka/node_modules/gulp/node_modules/chokidar/lib/fsevents-handler.js:292:17)
    at Array.forEach (native)
    at EventEmitter.FSWatcher.add (/Users/Adam/Sites/clients/mka/node_modules/gulp/node_modules/chokidar/index.js:420:11)
    at Object.exports.watch (/Users/Adam/Sites/clients/mka/node_modules/gulp/node_modules/chokidar/index.js:508:33)
    at Gulp.watch (/Users/Adam/Sites/clients/mka/node_modules/gulp/index.js:38:26)
    at watchClientCode (/Users/Adam/Sites/clients/mka/gulpfile.js:714:8)
    at bound (domain.js:250:14)
    at runBound (domain.js:263:12)
    at asyncRunner (/Users/Adam/Sites/clients/mka/node_modules/gulp/node_modules/undertaker/node_modules/bach/node_modules/async-done/index.js:38:18)
```

The path provided is actually an array, coming from gulp watch:

```
[ 'client/app/**/*.js',
  'client/other/**/*.js',
  '!client/app/**/*.spec.js' ]
```

Is this no longer supported, or is it a bug with gulp watch?

es128:
Also adds an explicit error message when bad (non-string or array of strings) input is provided

Resolves #360

### DIFF
--- a/index.js
+++ b/index.js
@@ -8,6 +8,7 @@ var globparent = require('glob-parent');
 var isglob = require('is-glob');
 var arrify = require('arrify');
 var isAbsolute = require('path-is-absolute');
+var flatten = require('lodash.flatten');
 
 var NodeFsHandler = require('./lib/nodefs-handler');
 var FsEventsHandler = require('./lib/fsevents-handler');
@@ -471,7 +472,7 @@ FSWatcher.prototype._remove = function(directory, item) {
 FSWatcher.prototype.add = function(paths, _origAdd, _internal) {
   var cwd = this.options.cwd;
   this.closed = false;
-  paths = arrify(paths);
+  paths = flatten(arrify(paths));
 
   if (cwd) paths = paths.map(function(path) {
     if (isAbsolute(path)) {
@@ -529,7 +530,7 @@ FSWatcher.prototype.add = function(paths, _origAdd, _internal) {
 // Returns instance of FSWatcher for chaining.
 FSWatcher.prototype.unwatch = function(paths) {
   if (this.closed) return this;
-  paths = arrify(paths);
+  paths = flatten(arrify(paths));
 
   paths.forEach(function(path) {
     if (this._closers[path]) {
--- a/index.js
+++ b/index.js
@@ -474,6 +474,10 @@ FSWatcher.prototype.add = function(paths, _origAdd, _internal) {
   this.closed = false;
   paths = flatten(arrify(paths));
 
+  if (!paths.every(isString)) {
+    throw new TypeError('Non-string provided as watch path');
+  }
+
   if (cwd) paths = paths.map(function(path) {
     if (isAbsolute(path)) {
       return path;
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -8,6 +8,8 @@ var globparent = require('glob-parent');
 var isglob = require('is-glob');
 var arrify = require('arrify');
 var isAbsolute = require('path-is-absolute');
+var flatten = require('lodash.flatten');
+var isString = require('is-string');
```
paulmillr: do we really need a separate module for `.filter(item => typeof item === 'string')`?

paulmillr:
What if we declare flatten, arrify and is-string as three small separate functions inside the main js file?

That would reduce the number of dependencies used by chokidar. I think this is useful esp for our huge user base

es128:
I think we're better off using externally maintained tiny dependencies than redefining them within the lib.

These have way less impact as deps than 1) fsevents, and 2) all the stuff that comes with glob matching.

paulmillr:
looks good other than that

es128:
Actually turns out I broke the test... fixing/rebasing now

paulmillr:
> These have way less impact

would be great to see some numbers related to this statement on a big scale. 3 deps = at least 6-12 more https requests.

let's merge this right now though, we can do a research later

es128:
don't merge quite yet, let me turn the CI checks green

es128:
Regarding the deps, I tend to agree with [@sindresorhus' perspective](https://github.com/sindresorhus/ama/issues/10#issuecomment-117766328)

If the size of the dep tree were a major concern, we'd make all sorts of choices differently, like anymatch using micromatch vs minimatch. But I don't think it should be thought of that way. We use modules for any common repeatable utility, even if simple, and just stick to just what we use, avoiding any mega kitchen sink libs.

That said, yeah, maybe is-string in particular is too bloated for what we need here. I guess I'll replace it with a simpler inline implementation.

phated:
is-string seems overkill

es128:
okok it's gone now

index.js
```diff
@@ -471,7 +472,15 @@ FSWatcher.prototype._remove = function(directory, item) {
 FSWatcher.prototype.add = function(paths, _origAdd, _internal) {
   var cwd = this.options.cwd;
   this.closed = false;
-  paths = arrify(paths);
+  paths = flatten(arrify(paths));
+
+  function isString(maybeString) {
```
paulmillr: this would declare it on every `add` -- more memory usage etc.

paulmillr: let's move it out of the fn

es128: fixed 464e85d

phated:
@es128 @paulmillr will 1.2 be released today?

es128:
yup, actually i'll take care of that now

PR #368, opened 2015-10-16
:memo: elaborate further on initial `add` events triggering

basarat:
Not sure if this is by design or a bug. If its by design would be good to have this in the documentation. Otherwise I can create a bug :rose:

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,7 +156,7 @@ time with two arguments (the path and the
 [`fs.Stats`](http://nodejs.org/api/fs.html#fs_class_fs_stats)
 object of that path).
 * `ignoreInitial` (default: `false`). Indicates whether chokidar
-should ignore the initial `add` events or not.
+should ignore the initial `add` events (emitted when the file is added to disk or to the tracked files).
 * `followSymlinks` (default: `true`). When `false`, only the
 symlinks themselves will be watched for changes instead of following
 the link references and bubbling events through the link's path.


es128:
I don't understand the point you're trying to make with this change.

`ignoreInitial` has little to do with when files are added to disk. The point is that by default `add` events are emitted for all the files that are already there. This can be useful for knowing what's on disk being watched without having to separately use `node-glob` or something like that. Setting `ignoreInitial: true` suppresses that behavior if you don't care what the file tree looks like when watching starts.

es128:
Closing for now, but we can discuss further.

basarat:
> The point is that by default add events are emitted for all the files that are already there

I thought `add` events were about the inverse of `unlink` events ... i.e. a new file is added. I got that impression because of how they are listed (https://github.com/paulmillr/chokidar#getting-started) : 

image

**Question : Is there an event that gets emitted when a file is added to disk?**

Thanks for the clarification. I'll try and write something up :rose:

paulmillr:
> Is there an event that gets emitted when a file is added to disk

That's `add` when `ignoreInitial` is `true.

es128:
also after the `ready` event has been emitted regardless of the `ignoreInitial` setting, any `add` events represent a file newly written to disk

es128:
https://github.com/paulmillr/chokidar/commit/946ae862ff9d82076e59494ad16d765060290618 hopefully makes it a little clearer

PR #378, opened 2015-10-29
Recreating deleted directory doesn't trigger event

erezarnon:
If we add a watch on directory A, mkdir A/B, rm -r A/B, mkdir A/B,
mkdir A/B/C, no event comes from fs.watch on creation of A/B/C

Encountered on Ubuntu 14.04.3 LTS, Linux 3.16.0-50-generic x86_64

### DIFF
--- a/index.js
+++ b/index.js
@@ -453,6 +453,11 @@ FSWatcher.prototype._remove = function(directory, item) {
     return;
   }
 
+  // Avoid conflicts if we later create another directory with the same name
+  if (isDirectory) {
+    this.unwatch(path);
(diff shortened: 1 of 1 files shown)


paulmillr:
Good catch. Could you fix tests?

es128:
This fix doesn't make sense to me. Look at what the `_remove` and `unwatch` methods do, and see what's already happening a few lines higher up than this patch.

What we need first is a failing unit test for the buggy scenario described so that we can work on a fix (that doesn't introduce new regressions).

Issue #379, opened 2015-10-29
Issue with watching a deleted, then recreated directory


es128:
Opening an issue to describe the scenario presented in #378 

> If we add a watch on directory A, mkdir A/B, rm -r A/B, mkdir A/B,
mkdir A/B/C, no event comes from fs.watch on creation of A/B/C

> Encountered on Ubuntu 14.04.3 LTS, Linux 3.16.0-50-generic x86_64

es128:
Fixed by #380

rottmann:
Problem exists again in chokidar 3.2.2.
Linux Mint 19.2, Kernel 4.15.0-65-generic

PR #382, opened 2015-11-02
When adding a previously-deleted file, emit 'add', not 'change'

erezarnon:
Turns out #379 is an issue for files as well as directories, but manifests differently.

### DIFF
--- a/index.js
+++ b/index.js
@@ -460,8 +460,8 @@ FSWatcher.prototype._remove = function(directory, item) {
   var eventName = isDirectory ? 'unlinkDir' : 'unlink';
   if (wasTracked && !this._isIgnored(path)) this._emit(eventName, path);
 
-  // Avoid conflicts if we later create another directory with the same name
-  if (isDirectory && !this.options.usePolling) {
+  // Avoid conflicts if we later create another file with the same name
+  if (!this.options.useFsEvents) {
     this.unwatch(path);
   }
 };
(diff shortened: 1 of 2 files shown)


erezarnon:
I've fixed the issue on linux. My test fails on osx, but I can't fix it because I don't have a mac. First version of this pull request was with commit https://github.com/erezarnon/chokidar/commit/db6f928, which failed even more osx tests than the current commit. I hoped maybe the issue didn't exist on osx, so tried to fix with c5a3c2e by checking for !this.options.useFsEvents. But now the issue still exists on osx.

es128:
This scenario (almost) is already tested. https://github.com/paulmillr/chokidar/blob/master/test.js#L352

If a file is unlinked and re-added in quick succession, it is represented as a `change` event deliberately by default (with `atomic: true`).

So perhaps basing the change on `if (!this.options.atomic)` would be the appropriate approach.

erezarnon:
The mapping of add->change still works. If we base it on this.options.atomic, the bug will exist when there's a delay between unlink and add. Also, I think Chokidar should understand internally that the event is file creation, even if it will map the event before emitting it.

Either way, this would still fail the test cases, right?

es128:
Ok, right. I'll need to investigate the Mac situation.

es128:
Thanks! Landed as 5a6315ff47b04467e27f45f4434176578437b971

Issue #384, opened 2015-11-03
Events lost with awaitWriteFinish


nono:
Hi,

on node 0.10 or 0.12 and linux, with this program:

```coffee
#!/usr/bin/env coffee

chokidar = require 'chokidar'
path = require 'path'

watcher = chokidar.watch '.',
    cwd: path.resolve '.'
    atomic: true
    awaitWriteFinish:
        pollInterval: 100
        stabilityThreshold: 2000
    interval: 1000
    binaryInterval: 2000

watcher.on 'add', (file) ->
    console.log "#{file} was added"

watcher.on 'raw', (event) ->
    console.log 'raw', event
```

If I do these commands:

```
$ cp ~/Desktop/happycloud.png .
$ rm happycloud.png            
$ cp ~/Desktop/happycloud.png .
$ rm happycloud.png            
$ cp ~/Desktop/happycloud.png .          
$ rm happycloud.png            
$ cp ~/Desktop/happycloud.png .
$ rm happycloud.png            
$ cp ~/Desktop/happycloud.png .
$ rm happycloud.png            
$ cp ~/Desktop/happycloud.png .
```

The first events are reported, but after some time, only the raw events are fired:

```
[...]
happycloud.png was added
raw rename
raw rename
raw change
happycloud.png was added
raw rename
raw rename
raw change
happycloud.png was added
raw rename
raw rename
raw change
raw change
raw change
raw change
happycloud.png was added
raw rename
raw rename
raw change
raw rename
raw rename
raw change
raw rename
raw rename
raw change
raw rename
raw rename
raw change
raw rename
raw rename
raw change
```

nono:
It happens if I remove the file in the 2s after copying it (stabilityThreshold). Then, `this._pendingWrites` has an entry for `happycloud.png` and blocks all events here: https://github.com/paulmillr/chokidar/blob/master/index.js#L127

es128:
My quickie attempt at a fix for this didn't pan out.

@alexpusch would appreciate if you could take a look

alexpusch:
By design, when awaitWriteFinish is activated the 'add' event is held back until the file is fully written. If the file is deleted during the writing process, no event will be raised. In such case there is no 'add' event so it's illogical to raise an 'unlink' event.

es128:
@nono if the last action taken is to copy the file, does the `add` event appear?

nono:
@es128 exactly, in the last action, the `add` is not fired.

[...]

PR #385, opened 2015-11-03
Prevent stale this._pendingWrites entries

es128:
A quick attempt at resolving #384 (without reproducing or confirming the fix)

@nono could you please try this branch and let us know what impact it had on your use case?

// @alexpusch


nono:
@es128 I still have the bug with this branch

Issue #381, opened 2015-11-02
awaitWriteFinish not taking effect on file changes


ng-dan:
I have the following code:
```JavaScript
var watcher = require('chokidar');
var watch = watcher.watch('file,dir or glob', {
        persistent: true
    });

watch.on('change', function (path, stats) {
    if (stats && stats.size > 0) {
        console.log("Reading file with size: " + stats.size);
    }
}
```

However when I upload a file (from WebStorm) to where it is watching it will trigger the change event multiple times as the file is being uploaded, like so (the output from one file upload):
```
Reading file with size: 8192
Reading file with size: 16384
Reading file with size: 24576
Reading file with size: 32768
Reading file with size: 40960
Reading file with size: 49152
Reading file with size: 57344
Reading file with size: 65536
Reading file with size: 73728
Reading file with size: 74094
```

I've tried the usePolling option as well as the awaitWriteFinish however it does not help to fix this issue.

Any help on the matter would be appreciated.

es128:
This is specifically what `awaitWriteFinish` was meant to provide a solution to. Without it, this is expected behavior. 

Perhaps you need to tweak its poll interval to be even longer for this use case.

// @alexpusch

alexpusch:
What OS? What is the method of the upload? Ftp? If the upload is slow you might need to increase the awaitWriteFinish.stabilityThreshold option

ng-dan:
This is on a server running Debian 8, and I'm using SFTP in WebStorm to upload the file. I'll try tweaking the stabilityThreshold option and see what affect that has.

ng-dan:
I've changed it to include the awaitWriteFinish option however it hasn't made a difference - it still does the exact same thing. Perhaps I've done something wrong in the following?

```JavaScript
var watch = watcher.watch('file,dir or glob', {
        persistent: true,
        awaitWriteFinish: {
            stabilityThreshold: 60000,
            pollInterval: 100
        }
    });
```

es128:
If it hasn't made any difference then you most likely don't have chokidar 1.2.0 installed.

ng-dan:
I've update chokidar to 1.2.0, but it's still doing the same thing when it comes to a file being uploaded.

alexpusch:
How big are the uploaded files? How fast is the upload? I can't think of a reason why an SFTP upload would change anything in the process.

ng-dan:
The files aren't big at all, around 73Kb and it uploads in less than a minute.

[...]

PR #389, opened 2015-11-04
Improve stabilityThreshold

nono:
See #381


es128:
Makes sense to me.

@alexpusch wdyt?

es128:
The failures in Travis appear to only be because of bugs in coveralls that don't seem to be getting resolved by re-running the jobs.

All the tests passed.

alexpusch:
Well, @nano is right. One of those bugs that makes you wonder how it ever worked in the first place

PR #391, opened 2015-11-05
hack to fix #390

### LINKED ISSUE #390: Change event fires twice for the same file with a different path
On `OSX@10.11.1`. Using `chokidar@1.0.0-1.2.0`.

```
change src/assets/index.scss <= relative path from cwd
change /Users/james/code/github/generator-fed/demo/src/assets/index.scss
```

---

I found the `cwd` option which makes the path relative and fixes the issue, however chokidar should probably only emit one event regardless of whether the `cwd` option is set or not?

jameslnewell:


es128:
Right. Thanks for the attempt, but sorry this is not going to be the solution as it would introduce unexpected behavior.

The most useful place to start would be writing a test that demonstrates the issue.

PR #392, opened 2015-11-06
Make awf and cwd options compatibles

nono:
Before this commit, when `awaitWriteFinish` and `cwd` options were enabled at the same time and `cwd` was not the current directory, no `add` events were fired. It was because the `fs.stat` failed in `awaitWriteFinish`.

### DIFF
--- a/test.js
+++ b/test.js
@@ -1392,6 +1392,21 @@ function runTests(options) {
             })();
           });
       });
+      it('should be compatible with the cwd option', function(done) {
+        var spy = sinon.spy();
+        var testPath = getFixturePath('awf_cwd/awf_cwd.txt');
+        options.cwd = sysPath.dirname(testPath);
+        fs.mkdirSync(options.cwd);
+        stdWatcher()
+          .on('all', spy)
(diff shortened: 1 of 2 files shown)


test.js
```diff
@@ -1392,6 +1392,21 @@ function runTests(options) {
             })();
           });
       });
+      it('should be compatible with the cwd option', function(done) {
+        var spy = sinon.spy();
+        var testPath = getFixturePath('awf_cwd/awf_cwd.txt');
```
es128: When introducing a new fixture file to the tests it needs to be added to the `clean` function at the top. Or you could change this to reuse paths that are already on there.

This is likely what caused the tests to explode.

nono:
@es128 yes, it should be green now.

es128:
The failures are the same problematic test that's failing regardless of this PR. Haven't solved it yet, but anyway - doesn't need to hold this one up

> 1) chokidar polling watch options awaitWriteFinish should not raise any event for a file that was deleted before fully written:

Thanks for catching this issue and providing a test & fix!

PR #393, opened 2015-11-07
Fix awaitWriteFinish unlink issue

es128:
Looks like I finally squished this one.

// @paulmillr @alexpusch @nono

### DIFF
--- a/index.js
+++ b/index.js
@@ -243,7 +243,9 @@ FSWatcher.prototype._awaitWriteFinish = function(path, threshold, callback) {
         // prevent `unlink` race condition
         this._getWatchedDir(sysPath.dirname(path))
           .remove(sysPath.basename(path));
-        this._pendingWrites[path].cancelWait();
+        if (this._pendingWrites[path]) {  
+          this._pendingWrites[path].cancelWait();
+        }
 
         if (err.code == 'ENOENT') return;
         return callback(err);
--- a/index.js
+++ b/index.js
@@ -243,12 +243,10 @@ FSWatcher.prototype._awaitWriteFinish = function(path, threshold, callback) {
         // prevent `unlink` race condition
         this._getWatchedDir(sysPath.dirname(path))
           .remove(sysPath.basename(path));
-        if (this._pendingWrites[path]) {  
-          this._pendingWrites[path].cancelWait();
+        if (this._pendingWrites[path]) {
+          return this._pendingWrites[path].cancelWait();
         }
-
-        if (err.code == 'ENOENT') return;
-        return callback(err);
+        return;
       }
 
       var now = new Date();
--- a/index.js
+++ b/index.js
@@ -243,9 +243,8 @@ FSWatcher.prototype._awaitWriteFinish = function(path, threshold, callback) {
         // prevent `unlink` race condition
         this._getWatchedDir(sysPath.dirname(path))
           .remove(sysPath.basename(path));
-        if (this._pendingWrites[path]) {
-          return this._pendingWrites[path].cancelWait();
-        }
+        if (this._pendingWrites[path]) this._pendingWrites[path].cancelWait();
+        if (err.code === 'ENOENT') return callback();
         return callback(err);
       }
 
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -240,13 +240,19 @@ FSWatcher.prototype._awaitWriteFinish = function(path, threshold, callback) {
   (function awaitWriteFinish (prevStat) {
     fs.stat(fullPath, function(err, curStat) {
       if (err) {
-        delete this._pendingWrites[path];
+        // prevent `unlink` race condition
+        this._getWatchedDir(sysPath.dirname(path))
+          .remove(sysPath.basename(path));
+        if (this._pendingWrites[path]) {  
+          this._pendingWrites[path].cancelWait();
```
alexpusch: cancelWait calls callback(), and the the end of this branch calls callback(err). This will surly cause some inconsistencies.
With this code the 
```javascript
if (err.code == 'ENOENT') return;
```
is not needed.

es128: Ok, let's try it. 0274e07

es128: No, the uncaught `ENOENT` is happening in Travis... going to add that back, but higher up

es128: Actually, I don't see a reason for `cancelWait` to call the callback anyway. It doesn't seem like it would make anything happen.

alexpusch: So we have a race condition between the file polling and the unlink event. one of them should cancel every thing about the file - the (maybe) ongoing timeout, and release the pending [async function](https://github.com/paulmillr/chokidar/blob/master/index.js#L156). In my original code I've relayed on that the unlink event will cancel the wait, and if the polling will come first, it will notice that the file is missing (ENOENT) and do nothing. 

Have you seen a case where the unlink event does not arrive at all?

es128: The issue I observed in the original code that was causing tests to fail regularly was that `unlink` events were being emitted without any prior `add` event nor an entry in `_pendingWrites`. This happened because of the slowness of `fs.stat` calls, causing misalignment between `this._watched` tracking and `this._pendingWrites` (the file shows up in the former, but awf suppressed the corresponding `add` event from being emitted, then never tracked it because it was gone by the time of the `stat`).

Hmm, hard to explain.

alexpusch: Hmm... So maybe the solution is to add the path to _pendingWrites before calling fs.stat? this way, if an unlink event will arrive, it could cancel the wait successfully. I'll try it now

es128: I did try that. Didn't solve it on its own. The last change in this PR seems to be working

es128:
For some reason node 0.10 on Windows is very reluctant to cooperate, but tests are now green everywhere else. 

@alexpusch any concerns with the latest diff?

index.js
```diff
@@ -240,36 +240,40 @@ FSWatcher.prototype._awaitWriteFinish = function(path, threshold, callback) {
...
       if (err) {
-        delete this._pendingWrites[path];
-        if (err.code == 'ENOENT') return;
-        return callback(err);
+        // prevent `unlink` race condition
+        this._getWatchedDir(sysPath.dirname(path))
+          .remove(sysPath.basename(path));
+        if (this._pendingWrites[path]) this._pendingWrites[path].cancelWait();
+        if (err.code !== 'ENOENT') callback(err);
+        return;
```
alexpusch: If the file is deleted (err.code === 'ENOENT') callback will never be called.

[Here](https://github.com/alexpusch/chokidar/commit/e7c33696bd246adb39b341f80c80cf2859692a76) is my try for this fix. I've extracted the creation of _pendingWrites[path] outside of the polling function, and returned the task of deleting the entry to the unlink event (via cancelWait). [everything is green](https://travis-ci.org/alexpusch/chokidar)

es128: The callback doesn't necessarily need to be called to do what it currently does. It only emits either an error or the event, and sometimes we want neither. I agree that it's ugly though, and as a next step I wanted to look into refactoring further to deal with cleaning up this control flow. 

Please open your version in a PR so we can see how it fares on appveyor, then we can determine which direction seems better.

es128:
Resolved the issue with e7c33696bd246adb39b341f80c80cf2859692a76

PR #394, opened 2015-11-07
Handle awf unlink and polling race condition

alexpusch:


index.js
```diff
@@ -275,7 +263,22 @@ FSWatcher.prototype._awaitWriteFinish = function(path, threshold, callback) {
...
-  }.bind(this))();
+  }.bind(this));
+
+  if (this._pendingWrites[path] === undefined) {
+    this._pendingWrites[path] = {
+      lastChange: now,
+      cancelWait: function() {
+        delete this._pendingWrites[path];
+        clearTimeout(timeoutHandler);
+        return callback();
```
es128: why?

es128:
I like this change, but the inconsistency about treating the emitting function like an async-flow-pattern callback remain. I refactored a bit further in https://github.com/paulmillr/chokidar/tree/alexpusch-awf-fix-refactor in an attempt to make it clearer (to avoid confusion that if callback isn't called it would disrupt the necessary control flow)

index.js
alexpusch: So that the pending [function](https://github.com/alexpusch/chokidar/blob/fix/waitWriteFinishTests/index.js#L156) will be invoked. I know it won't do anything since err and stat are null, but I still think we should not allow it to just hang there

alexpusch:
Looks great, change in terminology achieved exactly what you said. I'll update my PR

es128:
No need. Your commit is already there. Will just merge the branch to master.

Thanks for your help with this.

PR #396, opened 2015-11-09
Add a property based testing script

nono:
Hello,

I wanted to make some tests for `awaitWriteFinish` but it can take times to generate a lot of tests. I tried something called _property based testing_ to generate these tests for me (with the jsverify lib). The tests are green on my machine (no bug found) but I wonder if it something you are interested in. Maybe we can add it to travis and appveyor to catch more bugs.

What do you think about this?

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,8 +32,11 @@
   "devDependencies": {
     "chai": "^3.2.0",
     "coveralls": "^2.11.2",
+    "es6-promise": "^3.0.2",
     "istanbul": "^0.3.20",
+    "jsverify": "^0.7.1",
     "mocha": "^2.0.0",
+    "rimraf": "^2.4.3",
     "sinon": "^1.10.3",
     "sinon-chai": "^2.6.0"
   },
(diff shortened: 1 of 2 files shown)


es128:
It's interesting, but I'm not yet persuaded it brings us much of an advantage. Seems like a powerful concept for "unit tests", but that isn't really what we're doing to test chokidar because of its unique circumstances compared with typical modules. 

The tests actually act more as integration tests, involving the fairly flaky and unpredictable stack of low-level file system constructs under it. If we somehow mocked out the entire file system with predictable reactions, then we could perhaps have actual unit tests, but still that probably doesn't have much value here. The whole point of the library is to cover up the flakiness and inconsistencies, so you kind of want them present when testing.

Some ideas that I *have* become interested in regarding the tests are

* to give each test a separate subdirectory sandbox instead of testing the exact same paths over and over - sometimes getting odd results out of all that churn (otoh sometimes catching edge cases because of it)
* to explore separating file writing operations to be in a separate (child) process from the one doing the watching/detection
* with the first item as a prerequisite, give [AVA](https://github.com/sindresorhus/ava) a try as a replacement for mocha, see if we can run all the tests much faster in parallel


Feel free to expand upon the advantages you perceive in what you've done in this PR to help me "see the light"

PR #398, opened 2015-11-11
Don't apply awaitWriteFinish polling until after ready event

### LINKED ISSUE #397: ready vs initial add events with awf
Hi,

when the `awaitWriteFinish` option, the `ready` event is fired before the `add` events for files that already exist on the watched space. Do you think it's normal? I'd have expected `ready` to be fired only after all the `add` events.

es128:
Resolves #397

@nono @alexpusch


nono:
:+1:

alexpusch:
Cool

PR #399, opened 2015-11-20
Expand awaitWriteFinish to hold 'change' events for existing files

alexpusch:
Fix issue #381


es128:
Nice work. Thanks @alexpusch

PR #406, opened 2015-12-02
Do not filter out symlink paths found under glob patterns

### LINKED ISSUE #293: files specified by pattern inside symlinked dir
chokidar seems to not trigger events for files matched by a glob/pattern inside a symlinked dir.
File structure is:
```
./folder/symlinked/file.js
```
Working:
```js
  chokidar.watch('./folder/symlinked/file.js', {}).on('all', function(event, path) {
    console.log('trigger by chokidar', path);
  });
```
not working:
```js
  chokidar.watch('./folder/symlinked/*.js', {}).on('all', function(event, path) {
    console.log('trigger by chokidar', path);
  });
```

Am I missing something or this a bug?

What I really want to do is watch `./folder/**/*.js`. My `folder` contains dirs which are part symlinks and part regular dirs. That worked for regular dirs, but not for symlinks ..

es128:
Resolves #293 

@paulmillr @alexpusch

### DIFF
--- a/index.js
+++ b/index.js
@@ -350,16 +350,39 @@ var replacerRe = /^\.[\/\\]/;
 FSWatcher.prototype._getWatchHelpers = function(path, depth) {
   path = path.replace(replacerRe, '');
   var watchPath = depth || !isGlob(path) ? path : globParent(path);
+  var fullWatchPath = sysPath.resolve(watchPath);
   var hasGlob = watchPath !== path;
   var globFilter = hasGlob ? anymatch(path) : false;
+  var follow = this.options.followSymlinks;
+  var globSymlink = hasGlob && follow ? null : false;
+
+  var checkGlobSymlink = function(entry) {
+    // only need to resolve once
+    // first entry should always have entry.parentDir === ''
+    if (globSymlink == null) {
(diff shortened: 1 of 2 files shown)


laggingreflex:
This still doesn't seem to fix #293. It only works when watching `symlink/**/*` directly. It still doesn't work when watching `folder/**/*` and there's a `folder/symlink/file` change..

adityavohra7:
bump. I opened an [issue](https://github.com/kimmobrunfeldt/chokidar-cli/issues/28) on [chokidar-cli](https://github.com/kimmobrunfeldt/chokidar-cli) which seems to be symptomatic of the problem above. 

I'm watching ```bower_components/**/*.js``` which contains both regular directories, and symlinks to directories (created via ```bower link```). Changes to the linked directories aren't detected.

trainerbill:
+1.  I have a modules/ folder with symlinks in it and watching ./modules/**/*.js.  Using Gulp 4 with latest chokidar.  It does not work with matching a pattern in folder/symlink/pattern.  If I place the direct file path it works with symlink and if I remove the symlink and copy the directory in /modules folder it works.

PR #409, opened 2015-12-06
Added inherits to enable ES6 subclassing of FSWatcher

kmalakoff:
I've been using FSWatcher directly in a couple of places and needed to monkey patch it for subclassing in ES6.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "anymatch": "^1.3.0",
     "async-each": "^0.1.6",
     "glob-parent": "^2.0.0",
+    "inherits": "^2.0.1",
     "is-binary-path": "^1.0.0",
     "is-glob": "^2.0.0",
     "path-is-absolute": "^1.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -45,6 +45,7 @@
     "anymatch": "^1.3.0",
     "async-each": "^0.1.6",
     "glob-parent": "^2.0.0",
+    "inherits": "^2.0.1",
```
paulmillr: Let's use `require('util').inherits` — the same one as this module, except that it's not available for browsers.

kmalakoff: I am using chokidar in the browser using a virtual filesystem (fs-memory) so the fewer node dependencies, the better for me.

es128: > I am using chokidar in the browser using a virtual filesystem (fs-memory)

I'm intrigued. Could you elaborate more on this?

I would expect any in-memory fs shim that provides an `fs.watch` method to work fine on its own without all the problems involved in interfacing with a real file system. What is chokidar's role in this case?

PR #416, opened 2015-12-14
Fix the emitted stats when using awf

nono:
When using the awaitWriteFinish options, the stats were emitted for the file as it was first detected, not as it is when the event if fired.

### DIFF
--- a/index.js
+++ b/index.js
@@ -180,7 +180,11 @@ FSWatcher.prototype._emit = function(event, path, val1, val2, val3) {
         emitEvent();
       } else if (stats) {
         // if stats doesn't exist the file must have been deleted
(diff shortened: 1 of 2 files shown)


paulmillr:
Thanks!

nono:
@paulmillr do you think to release a new version of chokidar with this fix?

paulmillr:
No. Need to fix travis-ci.

nono:
But the build is green on master. For this PR, the build went red only because of a transient error: one job had an error while fetching code from github (https://travis-ci.org/paulmillr/chokidar/jobs/96740391).

paulmillr:
@nono pushed.

PR #437, opened 2016-02-06
Allow user to set debounce delay for atomic option

egasimus:
For me (on an i3 laptop with good old 5400rpm hard drive), `atomic` only works with a delay of 500ms. I've added a way to configure it by setting `atomic` to a number rather than `true`.


paulmillr:
docs please.

egasimus:
https://github.com/paulmillr/chokidar/pull/443

PR #451, opened 2016-03-15
Fix #450 - Don't lose 'unlink' events with awf+cwd

### LINKED ISSUE #450: Missed 'unlink' events with cwd + awf
Hello,

I've been tracking a bug in chokidar where some 'unlink' events are not fired. I'm pretty sure it's in chokidar code, as some events are fired in for a path and later the same path is no longer in `getWatched()` and no `unlink` event have been fired for it.

I've added some console.log in chokidar code to try to understand what happened. I think its cwd + awf + a file updated and then moved.

https://github.com/paulmillr/chokidar/blob/master/index.js#L516-L520
Here, `path` is absolute but `this._pendingWrites` has relative paths as keys. So, `cancelWait` is not called and `this._pendingWrites` still has an entry for the relative path associated to path when `_emit` is called.

https://github.com/paulmillr/chokidar/blob/master/index.js#L144 +
https://github.com/paulmillr/chokidar/blob/master/index.js#L151
Here, the path is set relative to cwd. Thus, it is found in `this._pendingWrites` and the emit is cancelled.

If I am right, to fix it, we have to transform `path` in a relative path before checking it in `cancelWait`. But it's not enough.

In theory, awf makes an `add` and an `unlink` events to cancel. But, here, it's a `change` and an `unlink`, so only the update should be canceled. Should we track if the first event of awf is an `add` or a `change` to and return it to `cancelWait`, in order to `return` or not here: https://github.com/paulmillr/chokidar/blob/master/index.js#L519?

nono:
See #450

### DIFF
--- a/test.js
+++ b/test.js
@@ -1548,15 +1548,14 @@ function runTests(baseopts) {
             stdWatcher()
               .on('all', spy)
               .on('ready', function() {
-                fs.writeFile(testPath, 'edit', simpleCb);
-                w(function() {
+                fs.writeFile(testPath, 'edit', w(function() {
                   fs.unlink(testPath, simpleCb);
                   waitFor([spy.withArgs('unlink')], function() {
                     spy.should.not.have.been.calledWith('change', filename);
                     spy.should.have.been.calledWith('unlink', filename);
                     done();
                   });
-                }, 10)();
+                }));
               });
           }));
         }));
--- a/test.js
+++ b/test.js
@@ -1538,6 +1538,28 @@ function runTests(baseopts) {
             done();
           });
       });
+      it('should emit an unlink event when a file is updated and deleted just after that', function(done) {
+        var spy = sinon.spy();
+        var testPath = getFixturePath('subdir/add.txt');
+        var filename = sysPath.basename(testPath);
+        options.cwd = sysPath.dirname(testPath);
+        fs.mkdir(options.cwd, w(function() {
+          fs.writeFile(testPath, 'hello', w(function() {
+            stdWatcher()
+              .on('all', spy)
+              .on('ready', function() {
+                fs.writeFile(testPath, 'edit', simpleCb);
+                w(function() {
+                  fs.unlink(testPath, simpleCb);
+                  waitFor([spy.withArgs('unlink')], function() {
+                    spy.should.have.been.calledWith('unlink', filename);
+                    done();
+                  });
+                }, 10)();
+              });
+          }));
+        }));
+      });
     });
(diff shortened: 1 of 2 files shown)


test.js
```diff
@@ -1538,6 +1538,29 @@ function runTests(baseopts) {
...
+        var testPath = getFixturePath('subdir/add.txt');
+        var filename = sysPath.basename(testPath);
+        options.cwd = sysPath.dirname(testPath);
+        fs.mkdir(options.cwd, w(function() {
+          fs.writeFile(testPath, 'hello', w(function() {
+            stdWatcher()
+              .on('all', spy)
+              .on('ready', function() {
+                fs.writeFile(testPath, 'edit', simpleCb);
+                w(function() {
```
es128: Shouldn't this be moved to be the callback after the `writeFile` above? Also, overriding the timeout to 10ms makes detecting the lack of a `change` event not very meaningful. The default time, or anything that gives it some time to react while still within the awf stability threshold would be better.

es128:
The new test passes on OS X even before the fix. Do you know why?

test.js
nono: > Shouldn't this be moved to be the callback after the writeFile above?

I built this test case by trying to mimic what I saw in the logs where the error was found, and there was a very short delay between the write and the delete. But I don't think it's relevant here. Doing as you said you should work too, and is probably simpler to understand.

> Also, overriding the timeout to 10ms makes detecting the lack of a change event not very meaningful.

Yes, you are right. The `change` event won't appear before the `unlink` one in this test, and should be tested later in the test.

> The default time, or anything that gives it some time to react while still within the awf stability threshold would be better.

I'm not sure to understand what you have in mind. You have a lot more experience than me at testing chokidar. So, feel free to refactor this test!

es128: You're using the `w()` wrapper and specifying your own timeout at 10ms to override the default. I'm suggesting to just leave it with the default as long as that works.

```js
                fs.writeFile(testPath, 'edit', w(function() {
                  fs.unlink(testPath, simpleCb);
                  waitFor([spy.withArgs('unlink')], function() {
                    spy.should.not.have.been.calledWith('change', filename);
                    spy.should.have.been.calledWith('unlink', filename);
                    done();
                  });
                }));
```

nono: Ok

nono:
No, I don't know. Maybe the path for the `unlink` event is already relative on OSX (and not absolute like on linux).

es128:
Looks good. Thanks!

PR #471, opened 2016-04-26
take new events into account for awaitWriteFinish

ijliebemij:
I was having issues when listening to a local path which is mounted as a samba share for others to copy files. The problem is that samba reserves the filesize for that transfer and the stats function returns always the full size even with 1% of the transfer completed.

I understand this solution might not be suitable for some other users where the file size is the key point. Probably another config option instead.

### DIFF
--- a/index.js
+++ b/index.js
@@ -299,19 +299,20 @@ FSWatcher.prototype._awaitWriteFinish = function(path, threshold, event, awfEmit
     }.bind(this));
   }.bind(this));
 
-  if (this._pendingWrites[path]) this._pendingWrites[path].cancelWait();
-
-  this._pendingWrites[path] = {
-    lastChange: now,
-    cancelWait: function() {
-      delete this._pendingWrites[path];
-      clearTimeout(timeoutHandler);
-    }.bind(this)
-  };
-  timeoutHandler = setTimeout(
-    awaitWriteFinish.bind(this),
-    this.options.awaitWriteFinish.pollInterval
-  );
+  if (!(path in this._pendingWrites)) {
+    this._pendingWrites[path] = {
+      lastChange: now,
+      cancelWait: function() {
+        delete this._pendingWrites[path];
+        clearTimeout(timeoutHandler);
+        return event;
+      }.bind(this)
+    };
+    timeoutHandler = setTimeout(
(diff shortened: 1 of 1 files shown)


es128:
Actually this seems reasonable to me. Can't think of why it would cause problems for anyone using `awaitWriteFinish`.

@alexpusch @nono any thoughts?

nono:
Looks good to me

ijliebemij:
Then, can we assume any file size change will always be preceded by an event? If that's the case, we can probably remove the fs.stat on the awaitWriteFinish function, or just the file size comparison in case you want to keep the error check. 

If any event on a file is considered and not just the filesize, we can even remove the pollInterval option, as any event could reset the timeout, and unless any other event resets the timeout, the emit will be triggered after the stabilityThreshold.

alexpusch:
@ijliebemij 
I've done a simple experiment. I copied a 2Gb file into a watched dir:
```js
fs = require('fs')

fs.watch('dir', function(event){
  console.log('event', event);
})

var checkFileSize = function(){
  fs.stat('dir/file', function(err, stats){
    if(stats){
      console.log('stats.size', stats.size)
    }
    setTimeout(checkFileSize, 100)
  })
}

checkFileSize()
```
Here are the result on OS X:
```
stats.size 0
event rename
stats.size 59768832
stats.size 97517568
stats.size 135266304

... more size changes

stats.size 1978662912
stats.size 2009071616
stats.size 2040332288
event rename
```

Results on Ubuntu 15.04 (On Virtual Box)

```
event change
event change
event change
stats.size 17174528
event change

... more event change

event change
stats.size 22040576
event change

... more event change

event change
stats.size 27394048
event change

... more event change

goes on like this
```

As we can see the change event triggering is dependent upon the platform, on OS X there is a single event for file addition, and write end (awf is not necessary), and on Ubuntu there seems to be a change event for each filesize change (dough I can't tell for sure)

Maybe there could be a strategy that relay only on fs events, but it will probably won't be very reliable. 

Anyway, I think the PR addition is good.

ijliebemij:
Nice test indeed. Ok, my assumption was not correct then. Current checks seem to be the most consistent.

PR #475, opened 2016-04-28
Fix addDir with cwd+alwaysStat on OSX

nono:


nono:
@es128 on a OSX laptop, the new test case always fails before the fix. With the fix, it fails about half the time because fsevents fire its events several seconds after the directory has been renamed (timeout). When I setup a longer timeout and a `waitFor` that waits for 10s instead of 3.5s, it works. So, the fix is right but I don't know how to test it properly.

es128:
Strange that fsevents is *that* slow to respond. Can you try adding the increased timeout setting on this one test for now so we can at least see it passing in Travis? Then see what we can do about it from there.

nono:
Well, in fact, the test has suceeded for https://travis-ci.org/paulmillr/chokidar/jobs/126323660 with fsevents. But it fails with polling. I haven't tested polling on the OSX laptop and I don't see why it fails.

es128:
Probably a simple timing issue - need to add some artificial delays between steps in the test, like a `w()` wrap around the `fs.rename` perhaps. Polling mode if finicky like that, makes it tough sometimes to keep the tests in balance.

nono:
Well, adding a `w()` fixes the test on the laptop. But it makes things worse on travis

es128:
Alright, well fixed locally is progress, I guess. Let me see if I can play with it and come up with anything.

nono:
@es128 did you find time to look at this PR? If so, I'm interested by what you found.

nono:
@es128 hello, do you have some news for me?

nono:
I close this PR in favor of #490.

PR #476, opened 2016-04-29
fixing depreciation warnings for node v6.0.0

yawetse:
updated optional dependency fsevents


es128:
This is not necessary on a fresh `npm install` or after an `npm update`.

Issue #458, opened 2016-04-03
awaitWriteFinish does not work properly (at least for 'change' events)


DavidBord:
This is pretty much my configuration:

```
    var options = {
      ignoreInitial: true, persistent: true, depth: 0,
      awaitWriteFinish: {
        stabilityThreshold: 2000,
        pollInterval: 1000
      }};
    var watcher = chokidar.watch('/home/ubuntu/workspace', options);

    var log = console.log.bind(console);
    watcher
      .on('change', function (path) {
        log('File', path, 'has been changed');
      });
```

These are the times in which the file changed:
`
[2016-04-03 07:44:03.550] [LOG]   Saving file:/home/ubuntu/workspace/tmp2.py
[2016-04-03 07:44:03.566] POST /saveFile 200 27 - ::ffff:10.0.0.126 - 15.983 ms
`

This is what is printed to console:
`
[2016-04-03 07:44:05.560] [LOG]   File /home/ubuntu/workspace/tmp2.py has been changed
[2016-04-03 07:44:05.560] [LOG]   File /home/ubuntu/workspace/tmp2.py has been changed
[2016-04-03 07:44:05.560] [LOG]   File /home/ubuntu/workspace/tmp2.py has been changed
[2016-04-03 07:44:05.561] [LOG]   File /home/ubuntu/workspace/tmp2.py has been changed
[2016-04-03 07:44:05.561] [LOG]   File /home/ubuntu/workspace/tmp2.py has been changed
[2016-04-03 07:44:05.561] [LOG]   File /home/ubuntu/workspace/tmp2.py has been changed
[2016-04-03 07:44:05.561] [LOG]   File /home/ubuntu/workspace/tmp2.py has been changed
[2016-04-03 07:44:05.562] [LOG]   File /home/ubuntu/workspace/tmp2.py has been changed
[2016-04-03 07:44:05.562] [LOG]   File /home/ubuntu/workspace/tmp2.py has been changed
[2016-04-03 07:44:05.562] [LOG]   File /home/ubuntu/workspace/tmp2.py has been changed
[2016-04-03 07:44:05.562] [LOG]   File /home/ubuntu/workspace/tmp2.py has been changed
[2016-04-03 07:44:05.562] [LOG]   File /home/ubuntu/workspace/tmp2.py has been changed
[2016-04-03 07:44:05.563] [LOG]   File /home/ubuntu/workspace/tmp2.py has been changed
[...]

PR #481, opened 2016-05-10
awf not to require ready event

vdiez:
I did this change because files were being taken even with the awf options set until the ready event was not triggered. I'm watching near 10k+ in some servers, so it takes a while to complete, and some of them might be in the middle of a transfer when I start the process. 

Is there a reason why the lack of the ready event prevents awf being taken into account? Right now I cannot think of any, but of course there are many cases different than mine.

### DIFF
--- a/index.js
+++ b/index.js
@@ -115,7 +115,6 @@ function FSWatcher(_opts) {
   this._emitReady = function() {
     if (++readyCalls >= this._readyCount) {
       this._emitReady = Function.prototype;
-      this._readyEmitted = true;
       // use process.nextTick to allow time for listener to be bound
       process.nextTick(this.emit.bind(this, 'ready'));
     }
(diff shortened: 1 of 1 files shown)


vdiez:
#458 might be related. Never confirmed by the user though

es128:
It's set up this way because of #397, so no I don't think we can resolve the issue you encountered quite so simply.

Seems we might need internal tracking to differentiate between `add` events originating the directory tree scan as opposed to those arising from the actual watch APIs. Hope we can find a way that isn't terribly complex to accomplish.

Closing this PR for now.

vdiez:
Ok I see the use case now for this. Whenever I have more time (couple of months from now) I'll try to look for a valid solution. Adding the initial files to the _readyCount should make it I guess. 

Thanks

Issue #482, opened 2016-05-14
Is there a reason for the 128 FSEvents watcher limit?


ashaffer:
It's forcing my project to fall back to polling for a lot of the files I need to watch, which is making it quite slow. I tried stubbing it out so that it always returns true, and it seems to work just fine. Is there a reason that limit needs to be there?

btw: Talking about the `canUse` in `lib/fsevents-handler.js`

es128:
Some history for why this was implemented at https://github.com/substack/watchify/issues/106. There's an absolute system limit on OS X (last I checked) of around 450 FSEvents across all processes. 128 was  a somewhat arbitrary limit set to account for a few instances of chokidar being able to run before tripping the `CarbonCore.framework` errors.

Since recursive watching is nearly free in terms of resource utilization under FSEvents, chokidar tries to consolidate those underlying watcher instances to common parents, even if as a user you're asking for lots of watcher instances on different paths. 

What does your watch config look like that you're managing to even hit the 128 limit to begin with? When you stub `canUse` to `return true`, can you also log out the count how many instances you're managing to reach?

ashaffer:
In my project at the moment it seems to make its way up to 224 listeners. This seems to be proportional to the number of modules I happen to have `npm link`'d at the time. Is there some reason that would cause extra watchers to be created?

es128:
> seems to make its way up to 224 listeners

So if the limit weren't in place and you were to run two instances of this process, or grow the project to the point where you make your way into the 400s, you'd end up in territory where catastrophic failures happen in your process(es) that chokidar could not do anything to help you resolve.

> number of modules I happen to have npm link'd at the time. Is there some reason that would cause extra watchers to be created?

Yes! The symlinks are being followed to their destination and a separate watcher being opened on each. If you were to explicitly add a watcher on the common parent of where the linked modules live on your file system - could be a separate watcher that you don't bind event listeners to - I believe you'd eliminate the problem you're observing.

Some ideas about a generalized fix:

* Make that 128 limit configurable. But I don't like this, it's a band-aid that would only help in fairly narrow circumstances. There's no working around the 400-something system-wide limit afaik. And it exposes a new config element to users that's hard to contextualize, and I think would often be found confusing and/or misused.
* Pick another arbitrary limit of fsevents instances (30?) at which point run some logic on the watched paths looking for common parents to reduce to - using the existing shared instance mechanisms. This is like an automated version of the workaround I suggested to you above.

[...]

PR #483, opened 2016-05-17
Consolidate common folders to avoid hitting the FSEvent limit

ashaffer:
@es128  Here's the PR for #482. Let me know if you want me to fix anything up or change it.

It does a great job on my project. Goes from 200+ instances to 19-35 (it's somewhat non-deterministic because the way the algorithm works, it depends on the order they accumulate in).

### DIFF
--- a/lib/fsevents-handler.js
+++ b/lib/fsevents-handler.js
@@ -38,7 +38,7 @@ function createFSEventsInstance(path, callback) {
 function setFSEventsListener(path, realPath, listener, rawEmitter) {
   var watchPath = sysPath.extname(path) ? sysPath.dirname(path) : path;
   var watchContainer;
-  var parentPath = parentOf(watchPath);
+  var parentPath = sysPath.dirname(watchPath);
 
   // If we've accumulated a substantial number of paths that
   // could have been consolidated by watching one directory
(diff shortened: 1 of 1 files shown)


lib/fsevents-handler.js
```diff
@@ -89,6 +98,33 @@ function setFSEventsListener(path, realPath, listener, rawEmitter) {
   };
 }
 
+var consolidateThreshhold = 10;
+
+// Decide whether or not we should start a new higher-level
+// parent watcher
+function couldConsolidate(path) {
+  var candidates = {};
```
paulmillr: Could you conform to our code style here please?

ashaffer: Do you want the threshold variable at the top?

Also I noticed `candidates` is no longer being used.

ashaffer:
Fixed style.

lib/fsevents-handler.js
```diff
@@ -89,6 +102,30 @@ function setFSEventsListener(path, realPath, listener, rawEmitter) {
...
+        return true;
+      }
+    }
+  }
+
+  return false;
+}
+
+// Find the parent of a given path
+function parentOf (path) {
```
paulmillr: sysPath.dirname`??

ashaffer: Fixed. Sorry, didn't realize that `dirname` worked for directories parents as well.

ashaffer:
Any chance we could get a patch version bump for this so I can start using it in watchify without maintaining a fork?

PR #490, opened 2016-05-31
Fix add for renamed dir

nono:
This is the PR #475 + a fix for files in a renamed directory. Chokidar was sending `unlink` events for the previous location of the files, but no `add` for the new location.

The tests are not ideal (sometimes green and sometimes red on my macbook). But I think it's just that fsevents can be slow, not a real bug.

### DIFF
--- a/index.js
+++ b/index.js
@@ -206,7 +206,9 @@ FSWatcher.prototype._emit = function(event, path, val1, val2, val3) {
     this.options.alwaysStat && val1 === undefined &&
     (event === 'add' || event === 'addDir' || event === 'change')
   ) {
-    fs.stat(path, function(error, stats) {
+    var fullPath = path;
+    if (this.options.cwd) fullPath = sysPath.join(this.options.cwd, path);
+    fs.stat(fullPath, function(error, stats) {
       // Suppress event when fs.stat fails, to avoid sending undefined 'stat'
       if (error || !stats) return;
 
--- a/index.js
+++ b/index.js
@@ -206,8 +206,7 @@ FSWatcher.prototype._emit = function(event, path, val1, val2, val3) {
     this.options.alwaysStat && val1 === undefined &&
     (event === 'add' || event === 'addDir' || event === 'change')
   ) {
-    var fullPath = path;
-    if (this.options.cwd) fullPath = sysPath.join(this.options.cwd, path);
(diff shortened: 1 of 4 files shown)


paulmillr:
Failing.

nono:
@paulmillr yes, it's a know issue from the #475 PR. You can look at the comments for more details. I hope that @es128 will find time to help me.

es128:
Sorry that I haven't been able to give this any real attention. Looks like the latest iteration is doing pretty well in Travis though. Will let it finish and then I'll take another look here and see whether I have any feedback.

nono:
Yes, it wasn't easy but I think the tests are OK.

index.js
```diff
@@ -206,7 +206,9 @@ FSWatcher.prototype._emit = function(event, path, val1, val2, val3) {
     this.options.alwaysStat && val1 === undefined &&
     (event === 'add' || event === 'addDir' || event === 'change')
   ) {
-    fs.stat(path, function(error, stats) {
+    var fullPath = path;
```
es128: Ok, now that the tests are passing, a bit of code style feedback.

Would prefer this condensed to

```js
var fullPath = this.options.cwd ? sysPath.join(this.options.cwd, path) : path;
```

package.json
```diff
@@ -32,6 +32,7 @@
   "devDependencies": {
     "chai": "^3.2.0",
     "coveralls": "^2.11.2",
+    "graceful-fs": "4.1.4",
```
es128: How persistent were the EMFILE errors you'd observed? Not something I tended to struggle with before, and I'd prefer to drop this unless it's really necessary.

es128:
Good work resolving the test issues - appreciate your persistence. Generally LGTM and seems like we're really close to being able to merge this.

package.json
nono: With OSX + node 0.10, the error happened every time on the before hook (https://github.com/paulmillr/chokidar/blob/master/test.js#L66-L68). I think this limit on the number of file descriptors may also be observed on other node versions if more tests are added to the suite.

es128: Ah, interesting. So yeah, this should be better than rolling our own workaround.

paulmillr:
Thank you.

PR #497, opened 2016-06-13
Add CHOKIDAR_USEPOLLING env variable detection

### LINKED ISSUE #496: ENV variable override for polling
In many cases, chokidar is used several layers down in a project... so it's difficult (if possible at all) to make the necessary adjustment and switch to polling. It would be great if an env variable override was added to help with this issue. Perhaps something like `CHOKIDAR_USEPOLLING=true`?

yaycmyk:
To override the chokidar instance configuration for an entire stack at once, regardless of how it is used / dependency depth.

Have to do string comparison because `process.env` always sets variables as strings.

Closes #496

### DIFF
--- a/index.js
+++ b/index.js
@@ -90,6 +90,18 @@ function FSWatcher(_opts) {
     opts.usePolling = process.platform === 'darwin';
   }
 
+  // Global override (useful for end-developers that need to force polling for all
+  // instances of chokidar, regardless of usage/dependency depth)
+  if (process.env.CHOKIDAR_USEPOLLING !== undefined) {
+    var envLower = process.env.CHOKIDAR_USEPOLLING.toLowerCase();
+
+    if (envLower === 'false' || envLower === '0') {
+      opts.usePolling = false;
+    } else if (envLower === 'true' || envLower === '1') {
+      opts.usePolling = true;
+    }
+  }
+
   // Editor atomic write normalization enabled by default with fs.watch
   if (undef('atomic')) opts.atomic = !opts.usePolling && !opts.useFsEvents;
   if (opts.atomic) this._pendingUnlinks = Object.create(null);
--- a/test.js
+++ b/test.js
@@ -1765,4 +1765,61 @@ function runTests(baseopts) {
       });
     });
   });
+  describe('env variable option override', function() {
+    describe('CHOKIDAR_USEPOLLING', function() {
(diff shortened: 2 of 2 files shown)


index.js
```diff
@@ -90,6 +90,12 @@ function FSWatcher(_opts) {
     opts.usePolling = process.platform === 'darwin';
   }
 
+  // Global override (useful for end-developers that need to force polling for all
+  // instances of chokidar, regardless of usage/dependency depth)
+  if (process.env.CHOKIDAR_USEPOLLING) {
```
paulmillr: we should make it support `0` etc, not just true

yaycmyk: `0` coerces to false. Not sure what you're looking for here. It will automatically work for anything that coerces to true, like `1`, any text, the literal word or boolean value `true`

yaycmyk: Oh I see what you're saying now, sorry. Yeah sure.

es128:
Your new test is failing in Travis

yaycmyk:
@es128 yeah, looks like sinon didn't like the stubbing I was trying to do. Just pushed an update.

es128:
The latest push that's still running at the moment is what I'm looking at. It fails when the tests set `usePolling: true` to begin with. Possibly should just drop that first assertion.

es128:
Also, please switch to `should` syntax to match the rest of the tests

yaycmyk:
@es128 Fixed, thanks!

yaycmyk:
Wat

yaycmyk:
@paulmillr @es128 It's happy now and I adjusted the implementation to be more specific.

There's a test that seems to be timing out in Node 0.10 that is unrelated.

index.js
```diff
@@ -90,6 +90,18 @@ function FSWatcher(_opts) {
     opts.usePolling = process.platform === 'darwin';
   }
 
+  // Global override (useful for end-developers that need to force polling for all
+  // instances of chokidar, regardless of usage/dependency depth)
+  if (process.env.CHOKIDAR_USEPOLLING !== undefined) {
+    var env_lower = process.env.CHOKIDAR_USEPOLLING.toLowerCase();
+
+    if (env_lower === 'false' || env_lower === '0') {
```
paulmillr: can you change his to `envLower`?

yaycmyk: Done

PR #505, opened 2016-06-30
Avoid calling _emitReady twice in case of a stat error

### LINKED ISSUE #338: Multiple "ready" events when watching an array of paths
I am currently receiving multiple "ready" events when chokidar is watching less than 200 files, but once the number grows larger I only get one.

I'm using the [following configuration](https://github.com/elastic/kibana/blob/master/src/cli/cluster/ClusterManager.js#L58-L68):
```js
chokidar.watch([
  'src/plugins',
  'src/server',
  'src/ui',
  'src/utils',
  'config',
  'installedPlugins'
], {
  cwd: fromRoot('.'),
  ignored: /[\\\/](node_modules|bower_components|public|__tests__)[\\\/]/,
});
```

gagern:
The `stat` just accounts for one ready count, namely the one for initial search.  In case of persistent watching, there will still be a watcher set up for the missing file, after which `_emitReady` will be called.

This fixes #338.

### DIFF
--- a/lib/fsevents-handler.js
+++ b/lib/fsevents-handler.js
@@ -327,7 +327,6 @@ function(path, transform, forceAdd, priorDepth) {
   // evaluate what is at the path we're being asked to watch
   fs[wh.statMethod](wh.watchPath, function(error, stats) {
     if (this._handleError(error) || this._isIgnored(wh.watchPath, stats)) {
-      this._emitReady();
       return this._emitReady();
     }
 


gagern:
Note: this issue was likely introduced in 5284735a73a04da4717f3ad15fa52532e66b40b9. Looks like a mistake, not a deliberate duplication of that call.

paulmillr:
cc @es128

es128:
I suspect the double call actually was deliberate, although I can't currently explain why and shame on me for not leaving a comment explaining why this odd looking code would be desirable. The whole methodology I came up with for counting traversed paths to decide when to emit `ready` - while it has worked well generally - seems brittle and could likely use some substantial refactoring.

Anyway, this PR is causing a test to fail (https://github.com/paulmillr/chokidar/blob/master/test.js#L1081-L1084) indicating that if a watch path also matches something provided in the `ignored` option that the `ready` event will fail to be emitted.

Perhaps that's a lesser evil than the one that this patch solved, but more effort will be needed to solve this, starting with a new test that specifically targets the #338 problem.

Issue #434, opened 2016-02-02
Node doesn't exit after closing watch


EzraM:
After setting up, then closing, a watcher, the Node process doesn't exit.
Windows.

``` javascript
const chokidar = require('chokidar')
const options = {}
const directoryPath = srcPath
// configuration taken from webpack
const watcher = chokidar.watch(directoryPath, {
    ignoreInitial: true,
    persistent: true,
    followSymlinks: false,
    depth: 0,
    atomic: false,
    alwaysStat: true,
    ignorePermissionErrors: true,
    usePolling: options.poll ? true : undefined,
    interval: typeof options.poll === "number" ? options.poll : undefined
});
watcher.close()
```

nicokaiser:
Maybe this has something to do with nested directories. We observed this with deep directories, but it worked with flat ones without subdirectories.

DarkPark:
I confirm the problem. No nested directories are required.
A simple code to reproduce: 

```js
var chokidar = require('chokidar');

var fn = function ( name ) {
    console.log('change: ' + name);
};

var watcher = chokidar.watch('./*')
    .on('change', fn)
    .on('unlink', fn)
    .on('add',    fn);

setTimeout(function () {
    console.log('should exit!');
    watcher.close();
}, 10000);
```

Interesting point to note - just run the code above and wait doing nothing - it will exit the node process in 10 seconds without any problems.
It will NOT exit in case there will be some file modifications in the current dir before `watcher.close`. So even once change/unlink/add are triggered and this hangs the process.

eoinmurray:
Same issue here -> quite important for us because we need many watchers to be opened and closed in a long running program, and close locks of the program

eoinmurray:
Its the `FSWatcher.unwatch` which is hanging, and I think this is being called inside `FSWatcher.close` 

```javascript
    watchers.forEach((w)=>{
      console.log(`unwatching ${watchers.length}`)
      const watched = keys(w.getWatched())
      console.log('gotWatched()')
      w.unwatch(watched)
      console.log('unwatched')
    })
  })
```

Fails sometimes on the `w.unwatch` step

es128:
Unfortunately it is most likely an issue with the underlying watch methods. I don't have high confidence that this is something that is possible to fix from within chokidar. Also likely to be highly OS-dependent. Would appreciate if anyone reporting this issue would note their OS & node version.

[...]

PR #536, opened 2016-10-02
Add ready-aware `.close` method

richardscarrott:
https://github.com/paulmillr/chokidar/issues/434

I'm not sure how you wanted to approach this but I initially made `.close` itself ready-aware however  it added a bunch of complexity to the test suite and slowed it down considerably which made me think it would be a pretty significant breaking change. You can see that attempt here -- https://github.com/60frames/chokidar/commit/192c6f6edc08100c5014d1de1a01f540d0236499

I then decided to revert back and take the easier approach of creating a separate `.stop` method, seen in this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,6 +238,7 @@ name and path for every event other than `ready`, `raw`, and `error`.
 * `.unwatch(path / paths)`: Stop watching files, directories, or glob patterns.
 Takes an array of strings or just one string.
 * `.close()`: Removes all listeners from watched files.
+* `.stop(?cb)`: Asynchronous `.close()`. (https://github.com/paulmillr/chokidar/issues/434)
 * `.getWatched()`: Returns an object representing all the paths on the file
 system being watched by this `FSWatcher` instance. The object's keys are all the
 directories (using absolute paths unless the `cwd` option was used), and the
--- a/index.js
+++ b/index.js
@@ -658,11 +658,21 @@ FSWatcher.prototype.unwatch = function(paths) {
   return this;
 };
 
-// Public method: Close watchers and remove all listeners from watched paths.
-
+// Private method: Close watchers and remove all listeners from watched paths.
+//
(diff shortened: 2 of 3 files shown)


es128:
Thanks for the PR. As much as I see how this would solve some problems, I would prefer if we went in the direction of making `.close()` smarter rather than making the API more confusing with slightly differentiated new methods.

Will not close this, though. Interested in other opinions, and I could be persuaded.

Thoughts on how else to approach this problem: Rather than actually waiting for `ready`, perhaps if `.close()` is called while the file scanning is still going on it could set an attribute that aborts any other watchers from being instantiated after that point. Additionally, the current method of holding onto the close/teardown methods from the underlying watch methods is clearly somewhat brittle (race conditions related to calling `.watch()` on new paths on an already-running chokidar instance are of particular concern)  and some attention could be paid there to make sure nothing is lost by the time `.close()` is called.

richardscarrott:
@es128 yeah I when I came to update the README it did start to feel pretty gross to be fair. If I get some time I'll see if I can work out something better based on your feedback.

rasenplanscher:
I think, there should definitely be only one such method. Why would anyone even want an *unaware* variant? Unless there was a noticable performance impact, in which case the solution would be questionable anyway.

Here's a question that popped into my head reading this: Are you guys clear on how this should work precisely, viewed from the outside? From the method name, I'd expect the `FSWatcher` to stop emitting events and to stop accepting new watches. The docs read like it's basically an `unwatch` on everything. This probably has implications as to how to do this...

es128:
@rasenplanscher the issue is that chokidar has a 2-step process. 1) walk the fs tree(s) being watched, 2) establish watchers. The current `.close()` does not pay any attention to step 1, so it fails if called before that is complete.

But the proper solution isn't to wait until step 1 is done. What's needed is to find a way to abort step 1 cleanly while it's in midstream.

es128:
Replaced by #600

Issue #447, opened 2016-03-09
Mac: High CPU usage for large folders


bpasero:
This is more a question than a bug report with concrete steps. VS Code (using chokidar 1.0.5) has numerous bug reports where Mac users report high CPU usage for our file watcher. The only thing in common is that these users open VS Code on a very large folder (> 2 GB) which seems to cause the high CPU spikes.

Is there any known issue for chokidar or fsevents where on a Mac CPU would go crazy when a folder is large? I would assume that the folder size has no impact on how chokidar/fsevents behaves but maybe there is something I am missing?

es128:
The physical size of the files shouldn't have an impact, but a deeply nested directory tree with many entries would have an impact on the initial scan of the file structure. It should subside after the `ready` event if the paths were passed with a new watcher instance (if using `.add()` the process is the same, but there is no similar `ready` signal when the scan is done in that case).

Also, using `usePolling: true` in such a situation would cause persistently high CPU utilization.

bpasero:
@es128 interesting, I was not aware of the initial scan, is there a reason why you need to scan all files on startup? does this scan ignore folders that are excluded with the exclude setting?

es128:
Skimmed through the linked issue. Switching versions of node with nvm *will* break `fsevents` and cause chokidar to silently revert to polling. The solution is to `npm rebuild` after any change in the version of node. This presumes that VS Code uses whichever version of node is in `$PATH` as opposed to locking down to a bundled version.

bpasero:
@es128 I think this issue is independent and only shows up when running Code from its source code form. When running the bundled product, the environment is made of the node environment Electron provides and is always stable independent from the node version installed.

es128:
You and I have discussed the initial scan with `readdirp` in the past. Chokidar needs a reference point for the file tree in order to correct erratic events that come from the lower-level watch methods, such as signals emitted on a directory when a file was added or removed. It does avoid recursing through `ignored` paths.

The paths it is keeping track of can now be exposed with the `.getWatched()` method.

Just FYI - there's been some movement toward making the scan more efficient (#412), but I'll need to focus more time and attention toward it at some point to work toward landing those improvements.

es128:
But I don't think the scan has much to do with the original issue report. [The symptoms described in this comment](https://github.com/Microsoft/vscode/issues/3222#issuecomment-194119602) make it pretty clear to me that for whatever reason polling mode is being used in that case.

bpasero:
[...]

PR #539, opened 2016-10-08
Improve ignore paths globifying

laggingreflex:
Ensures that if only a directory's basename is provided (without the preceeding path, just `dirname`), it prefixes `**/`

Currently when only a directory name is provided, it adds a postfix `/**` which is great for matching the entire tree ***if*** the tree is a *child* of the dir.  So `path` (converted to `path/**`) can filter paths like `path/to/file.js`, **but** if `path` is actually only a directory's basename it fails to match `/some/path/to/file.js`. This can be solved by prefixing `**/path`

Really helpful when only "node_modules" is provided as ignore path (like in `.gitignore` file). 

This could probably have solved #447, which if I followed correctly [the solution][1] ultimately was to add paths in similar patterns

        "**/.git/objects/**"
        "**/node_modules/**"

Also adjust the postfix to add just `**` instead of `/**` based on if last char is already a `/`

[1]: https://github.com/Microsoft/vscode/issues/3222#issuecomment-195001923

### DIFF
--- a/index.js
+++ b/index.js
@@ -34,6 +34,19 @@ var isString = function(thing) {
   return typeof thing === 'string';
 };
 
+var globify = function(path) {
+  var ret;
+  if (path.slice(-1) === sysPath.sep) {
+    ret = path + '**';
+  } else {
+    ret = path + sysPath.sep + '**';
+  }
+  if (!sysPath.isAbsolute(path)) {
+    ret = '**' + sysPath.sep + ret;
+  }
+  return ret;
+};
+
 // Public: Main class.
 // Watches files & directories for changes.
 //
@@ -356,9 +369,7 @@ FSWatcher.prototype._isIgnored = function(path, stats) {
     var paths = arrify(ignored)
       .filter(function(path) {
         return typeof path === 'string' && !isGlob(path);
-      }).map(function(path) {
-        return path + '/**';
-      });
+      }).map(globify);
     this._userIgnored = anymatch(
       this._globIgnored.concat(ignored).concat(paths)
     );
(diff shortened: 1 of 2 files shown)


es128:
At a glance, this doesn't make sense to me. The relative-ness of the glob paths should match with the way the paths are specified for being watched, and that's how it's output. If you watch with absolute paths, your glob patterns need to reflect the entire absolute path and vice versa.

If something about that isn't working as expected, we can look at it and see if we can solve for the underlying cause to make sure the paths being matched reflect the same basedir as the patterns, but automatically prefixing with `**` seems like an invitation for new bugs.

Maybe instead of prefixing with `**` we could consider prefixing any relative ignore paths with the glob-parent basedir of the watch paths to accomplish the goal of this PR.

laggingreflex:
You're right my bad. It seems it already works the way you described, I just needed to add `cwd` option.

laggingreflex:
So just to clarify, it only excludes either when both the watch path *and* exclude paths are either *both* relative or when both are absolute. I had a watch path as an absolute path and ignored paths as relative paths. So in this case it is required to have `cwd` option passed in for it to work, right?

Would it be desirable to have it auto infer `cwd` if the watch path is absolute? Or is this intended?

es128:
Better to think in terms of how globs are matched. The pattern `path/to/*.js` does not match on `/some/path/to/file.js` unless you first convert one or the other to be treated as absolute or relative from the same base. 

> Would it be desirable to have it auto infer cwd if the watch path is absolute? Or is this intended?

This is the same as the suggestion I made at the end of my comment. Might be worth it, but there might also be a risk of side effects. 

In your case using `cwd` is one viable solution. The other is to keep the inputs consistent with each other.

PR #554, opened 2016-11-18
glob-parent@3

antialias:
This should fix https://github.com/paulmillr/chokidar/issues/344 and possibly help with https://github.com/paulmillr/chokidar/issues/300

(I've only tested this on Mac OS X)


paulmillr:
seems like tests are failing

es128:
I'm still working on making glob-parent@3 ready for chokidar. It isn't yet. I'll make the update when it is.

Issue #518, opened 2016-08-03
Add environment support for configuring the poll interval


langri-sha:
Currently, `CHOKIDAR_USEPOLLING` can be set to either 0, 1 or an arbitrary value. Would it make sense to use values greater than 1 to configure the desired polling interval (or introduce another flag)? 

This would be immensely useful for environments where resources may be scarce, say a Docker container.

es128:
Probably its own flag for cleanliness. `CHOKIDAR_POLLINTERVAL` maybe.

PR welcome.

PR #557, opened 2016-11-27
Added ability to force `interval` value by setting `CHOKIDAR_INTERVAL…

YAmikep:
Trying to add the ability to force `interval` value by setting `CHOKIDAR_INTERVAL` env variable.
Ref: https://github.com/paulmillr/chokidar/issues/518

### DIFF
--- a/index.js
+++ b/index.js
@@ -106,8 +106,7 @@ function FSWatcher(_opts) {
   }
   var envInterval = process.env.CHOKIDAR_INTERVAL;
   if (envInterval) {
-    var envInteger = parseInt(envInterval);
-    opts.interval = envInteger;
+    opts.interval = parseInt(envInterval);
   }
 
   // Editor atomic write normalization enabled by default with fs.watch
--- a/index.js
+++ b/index.js
@@ -106,7 +106,8 @@ function FSWatcher(_opts) {
   }
   var envInterval = process.env.CHOKIDAR_INTERVAL;
   if (envInterval !== undefined) {
-      opts.interval = envInterval;
+    var envInteger = parseInt(envInterval);
+    opts.interval = envInteger;
   }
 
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -104,6 +104,11 @@ function FSWatcher(_opts) {
       opts.usePolling = !!envLower
     }
   }
+  var envInterval = process.env.CHOKIDAR_INTERVAL;
+  if (envInterval !== undefined) {
```
es128: Simply `if (envInterval)` should be fine in this case, since we're not contending with `false` as a valid potential value.

index.js
```diff
@@ -104,6 +104,11 @@ function FSWatcher(_opts) {
...
+    var envInteger = parseInt(envInterval);
+    opts.interval = envInteger;
```
es128: Combining these two lines to just `opts.interval = parseInt(envInterval)` should be fine.

YAmikep:
@es128 any idea why the build fails for oxs+node7 and osx+node4?

es128:
@YAmikep seems like Travis' OS X infrastructure is somewhat underpowered and has a hard time with the rapid disk operations in the tests. It's something I've struggled with for a while, typically just need to restart the failing jobs until they pass.

sxn:
Any update on this? Seems to have died off a bit with the holidays & all :)

AJamesPhillips:
Bump.  Anything I can do to help?  And thank you very much for all the work you've put in!  Really appreciate it 👍  :)

es128:
It's been a while, I know, but I managed to put some attention in to chokidar this weekend, and should be able to turn my attention to this PR quite soon.

I want to see the test suite pass consistently in CI, even though those issues are likely unrelated to this PR. I think I've gotten close to resolving it.

AJamesPhillips:
👍   Thanks @es128 :)

PR #586, opened 2017-03-14
🐜  Detect non-null 'options.ignored'

aleclarson:
The value of `options.ignored` can be a string, function, regexp, or mixed array.

Strangely, no tests failed before or after this change.
Probably need to add an edge case.


aleclarson:
Ack, looks like the `checkIgnored()` call handles the other cases.

Now I understand that `checkIgnored(stats)` is only necessary when `options.ignored` is a function.

But I wonder... what if an array of ignore functions are used?

paulmillr:
Closing for new architecture.

PR #587, opened 2017-03-14
🐜  Make 'ready' event work as expected

aleclarson:
- Add a missing `_emitReady` call when `options.depth` is exceeded.
- Only call `_emitReady` once after fetching the `stats` of a watched path.

---

### Failing tests

```
  1) chokidar fsevents (native extension) watch options ignored should not choke on an ignored watch path:
     Error: Timeout of 6000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

Looks like the "ready" event is not emitted when zero files are watched.


paulmillr:
Closing for new architecture.

PR #598, opened 2017-04-21
ignoreGlobs option: treat glob-like paths as literal paths

octachrome:
As discussed with @es128 at https://github.com/paulmillr/chokidar/issues/300.

Naming options is always tricky, so shout if you have a better suggestion than `ignoreGlobs`.

If you're happy with the code, I'll update the docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,7 @@ chokidar.watch('file', {
   ignoreInitial: false,
   followSymlinks: true,
   cwd: '.',
+  disableGlobbing: false,
 
   usePolling: true,
   interval: 100,
@@ -168,6 +169,8 @@ symlinks themselves will be watched for changes instead of following
 the link references and bubbling events through the link's path.
 * `cwd` (no default). The base directory from which watch `paths` are to be
 derived. Paths emitted with events will be relative to this.
+* `disableGlobbing` (default: `false`). If set to `true` then the strings passed to `.watch()` and `.add()` are treated as
+literal path names, even if they look like globs.
 
 #### Performance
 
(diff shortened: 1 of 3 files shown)


es128:
Cool, I'm glad the easy approach seems to work out. Thanks for the PR.

As for the option name, I don't think `ignoreGlobs` is very clear about the intent. Perhaps `disableGlobbing`?

Also, if you don't mind, can we add tests showing this working on an individual file, as well as showing that other files that *would* match the glob pattern (if it were being treated as such) do not get watched.

octachrome:
@es128 I added some more tests - do they address your concerns?

es128:
Yes. Thank you, great work!

DaleMckeown:
Would I be right in saying that this `ignoreGlobs` option affects the watcher path only and not any filepaths passed to the `ignored` option?

I can't seem to get something like this to work - tried with relative and absolte pathing. It always picks up the changes!

```
Chokidar.watch("../Folder (1)", { disableGlobbing: true, ignored: "bin", ignoreInitial: true, depth: 5 }).on('all', function (event, path) {
    console.log(event);
    console.log(path);
}).on('error', function (error) {
    console.log(error);
});
```

octachrome:
@DaleMckeown The docs for `ignored` say "The whole relative or absolute path is tested, not just filename". So you need to use a regex. If you want to ignore files in any subdirectory named bin:

    ignored: /\/bin\//,

Or if you only want to ignore files in a directory named bin in your root directory:

    ignored: /^\.\.\/Folder \(1\)\/bin\//,

DaleMckeown:
@octachrome Ah, thanks! I'll take a look and see if I can get that working tomorrow.

Issue #599, opened 2017-04-25
Show error message about why `fsevents` could not be required


vjpr:
At present the error is silenced:

```
try { fsevents = require('fsevents'); } catch (error) {}
```

https://github.com/paulmillr/chokidar/blob/85df8a78af3fd654784286cd79c9307391c18357/lib/fsevents-handler.js#L7

Sometimes `fsevents` cannot be required because it has been compiled against the wrong node abi. 

```
try { fsevents = require('fsevents'); } catch (error) { console.log(error) }
```

```
Error: Module did not self-register.
    at Error (native)
    at Object.Module._extensions..node (module.js:583:18)
    at Module.load (module.js:473:32)
    at tryModuleLoad (module.js:432:12)
    at Function.Module._load (module.js:424:3)
    at Module.require (module.js:483:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/Users/Vaughan/dev-live/webpackerator/node_modules/.registry.npmjs.org/fsevents/1.1.1/node_modules/fsevents/fsevents.js:14:14)
    at Module._compile (module.js:556:32)
    at Object.Module._extensions..js (module.js:565:10)
    at Module.load (module.js:473:32)
    at tryModuleLoad (module.js:432:12)
    at Function.Module._load (module.js:424:3)
    at Module.require (module.js:483:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/Users/Vaughan/dev-live/webpackerator/node_modules/.registry.npmjs.org/chokidar/1.6.1/node_modules/chokidar/lib/fsevents-handler.js:7:18)
```

Or the user has not installed it correctly.

We could check for the different kinds of errors that occur when requiring a module and show an informative error message.

A very common issue is high CPU usage (100%+) when `fsevents` fails to be required when using `webpack` (`watchpack` requires `chokidar`). I would say that a lot of the time the user doesn't notice what went wrong, but their computer is being slowed down immensely. For this reason, I )would argue for the error message to be shown...or have an event hook...or have an environment variable that would allow printing the error message. E.g. `process.env.CHOKIDAR_PRINT_FSEVENTS_STATUS`.

es128:
Not going to have chokidar start printing messages. That's the domain of the app/tool, not the utility module.

[...]

PR #602, opened 2017-05-08
add requireFseventsCallback option

vjpr:
First draft.

I wanted to make the smallest change possible.

I initialized `var fsevents = {}` to get around the `'use strict'` restrictions of modifying uninitialised variables.

I chose a callback over an emitter because some of the options are updated depending on the result of `FsEventsHandler.canUse()` which require `fsevents` to have attempted to be required.

Not sure on the name of the option. Maybe its better as `requireFseventsErrorCallback`.

No tests at present but wanted to get feedback on approach first.

#599 

@es128

### DIFF
--- a/index.js
+++ b/index.js
@@ -83,7 +83,7 @@ function FSWatcher(_opts) {
   if (undef('useFsEvents')) opts.useFsEvents = !opts.usePolling;
 
   if (opts.useFsEvents) {
-    FsEventsHandler.initFsEvents(opts.fseventsPath, opts.requireFseventsCallback);
+    FsEventsHandler.initFsEvents(opts.fseventsPath, opts.useFsEvents, opts.requireFseventsCallback);
   }
 
   // If we can't use fsevents, ensure the options reflect it's disabled.
--- a/index.js
+++ b/index.js
@@ -82,10 +82,16 @@ function FSWatcher(_opts) {
   // Enable fsevents on OS X when polling isn't explicitly enabled.
   if (undef('useFsEvents')) opts.useFsEvents = !opts.usePolling;
 
-  if (opts.useFsEvents) FsEventsHandler.init(opts.requireFseventsCallback);
+  if (opts.useFsEvents) {
+    FsEventsHandler.initFsEvents(opts.fseventsPath, opts.requireFseventsCallback);
+  }
 
   // If we can't use fsevents, ensure the options reflect it's disabled.
-  if (!FsEventsHandler.canUse()) opts.useFsEvents = false;
+  if (!FsEventsHandler.canUse()) {
+    opts.useFsEvents = false;
+  } else {
+    importHandler(FsEventsHandler);
+  }
 
   // Use polling on Mac if not using fsevents.
   // Other platforms use non-polling fs.watch.
@@ -702,7 +708,6 @@ function importHandler(handler) {
   });
 }
 importHandler(NodeFsHandler);
-if (FsEventsHandler.canUse()) importHandler(FsEventsHandler);
 
 // Export FSWatcher class
 exports.FSWatcher = FSWatcher;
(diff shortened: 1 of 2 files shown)


es128:
Yeah the name is clunky, but I don't have a better suggestion atm.

Approach LGTM.

index.js
```diff
@@ -82,6 +82,8 @@ function FSWatcher(_opts) {
   // Enable fsevents on OS X when polling isn't explicitly enabled.
   if (undef('useFsEvents')) opts.useFsEvents = !opts.usePolling;
 
+  FsEventsHandler.init(opts.requireFseventsCallback);
```
es128: May as well put `if (opts.useFsEvents)` in front of this. It will add the advantage of sparing some initialization time for everyone else.

vjpr:
Maybe renaming from `init` to `tryRequireFsEvents`? Or `tryInstantiateFsEvents`?

es128:
`initFsEvents`

vjpr:
`canUse` is used at the top-level scope, so maybe its better to initialise it in `canUse`?

Or I move the init function to here.

```
// Attach watch handler prototype methods
function importHandler(handler) {
  Object.keys(handler.prototype).forEach(function(method) {
    FSWatcher.prototype[method] = handler.prototype[method];
  });
}
importHandler(NodeFsHandler);
if (FsEventsHandler.canUse()) importHandler(FsEventsHandler);
```

vjpr:
- I have made the callback a Node `errback` which will report the resolve path of fsevents.
- I have added an option to configure the resolved path to fsevents. If a package is symlinked, it will not be able to find the root project's fsevents (with --preserve-symlinks). Allowing this option will mean consumers can have a way to manually locate fsevents.
- I have moved `importHandler` inside `FSWatcher`.

es128:
What just happened? Now it looks like kind of a mess. 

What was so wrong with where you had it before aside from initializing the `fsevents` var with an empty object instead of leaving it `undefined` or some other falsy value?

vjpr:
> What was so wrong with where you had it before aside from initializing the fsevents var with an empty object instead of leaving it undefined or some other falsy value?

See https://github.com/paulmillr/chokidar/pull/602#issuecomment-300009813.

vjpr:
I'm not happy with splitting up the `importHandler` methods. Maybe we should move the other `importHandler` into the constructor. Or we just leave the FsEventsHandler methods on the class and don't do the `canUse` check.

es128:
Sorry, I'm not understanding it right now. I'll need to try it locally when I get some time later to figure out why it couldn't have stayed simpler.

vjpr:
Without moving the `importHandler` it broke, because `FsEventsHandler.canUse` was being called in the top-level scope of `index.js` (when `require('chokidar')` is called) before opts are parsed.

Let me know if you would allow `opts.fseventsPath`. 

The rationale is: if you npm link to `webpack-dev-server` (for example) for debugging (or a package that requires `webpack-dev-server`), unless you `npm install fsevents` in `webpack-dev-server`'s dir, it will not use `fsevents`. And even if you do, your app would end up using two different instances of `fsevents`.

vjpr:
Just ran into this issue once again with a silent `Error: Module did not self-register.` message.

Its tough, because cpu spikes to 120%, which could be caused by something else in the app, but now I always have to find and then add console.log to that try/catch block.

With the current solution, if library devs (such as webpack-dev-server) don't implement the hook, then I am stuck, and we don't help anyone.

I think we really need an environment variable. I would set this env var permanently in my `~/.bashrc` on macOS so I could always get notified if fsevents wasn't used.

es128:
I haven't had time to review closely, but I'm sorry to tell you that I think as this PR grows it becomes less likely to be accepted. A note in the readme reminding you to `npm rebuild` if you switch node versions may be where we end up if we can't get to any other effective solution that is also simple and small.

I suggest you focus this down to the one simple solution that will have a positive impact on your quality of life relating to this problem, within the constraints I've already described. Perhaps that's going back to before the other changes you've made and only adding something `catch` block to print a message ONLY WHEN the user has set a specific env variable set. Yes, that only covers users savvy enough to set the env var, but I haven't seen any other proposal that won't cause collateral problems/complaints.

vjpr:
I am very happy with just the env var personally. I will put together a
fresh PR just for that.

PR #605, opened 2017-05-12
print fsevents require error when env var set

vjpr:
I have used the same env var parsing code as in index.js.

Question: I have used two `console.log` statements. Should it be combined into one perhaps?

Will add tests after approval.

#599

@es128

### DIFF
--- a/lib/fsevents-handler.js
+++ b/lib/fsevents-handler.js
@@ -3,33 +3,9 @@
 var fs = require('fs');
 var sysPath = require('path');
 var readdirp = require('readdirp');
-var fsevents = tryRequireFsEvents();
-
-function parseEnvVar(envVar) {
-  if (envVar !== undefined) {
-    var envLower = envVar.toLowerCase();
-    if (envLower === 'false' || envLower === '0') {
-      return false;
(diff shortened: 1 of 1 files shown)


lib/fsevents-handler.js
```diff
@@ -3,8 +3,34 @@
 var fs = require('fs');
 var sysPath = require('path');
 var readdirp = require('readdirp');
-var fsevents;
-try { fsevents = require('fsevents'); } catch (error) {}
```
es128: just keep it up here and leave it simple

```js
catch (error) {
  if (process.env.CHOKIDAR_PRINT_FSEVENTS_REQUIRE_ERROR) console.log(error)
}
```

Why do anything more than that? The rest is overkill.

vjpr: I am happy to do it like that :)

I only added the function because it was how it is done in `index.js`.

es128:
I see what you mean about the env parser.

In that case what we need is to DRY it up and make that function accessible to both, or move all env handling to one place.

I'd be fine accepting this simple PR and worrying about the refactoring separately.

What did you have in mind in terms of a test for this? Do you see a straightforward way to test it?

PR #616, opened 2017-06-20
Updates fsevents dependency

chrissrogers:
- fsevents is not publishing binaries for its older versions on new node versions.
- This updates fsevents so that package managers running on the latest node version will have binaries available


chrissrogers:
Misread the caret declaration -- dependency should be fine

PR #617, opened 2017-06-21
bump fsevents to 1.1

nueko:
`fsevents` version 1.1  successfully installed on windows but 1.0 not, with 1.1 `npm run test` passed.


es128:
Thanks for the PR, but this change is not necessary. If you were having problems on Windows, it was probably because of a shrinkwrap file generated on OS X that needs to be updated or otherwise fixed to be platform-independent.

Issue #287, opened 2015-04-19
Problem with braced glob patterns that match subdir contents, but not the subdir


ohjames:
Things like `build/{,test/}*.js` can sometimes be useful... when you have a project with lots of nested directories then using `build/**/*.js` can really slow things down, in other cases it's just a neat shorthand. Some other watch libraries support it but all of those have too many issues to be reliable alternatives to chokidar.

I'm happy to work on the code if you think this would be useful.

paulmillr:
Isn't this a micromatch issue?

ohjames:
micromatch supports expanded glob patterns already. I couldn't find `micromatch` in chokidar's dependency list, is it including it via a different package? Maybe that package needs to upgrade it's micromatch dependency or something?

es128:
micromatch is included via anymatch. All versions are at latest, although there are some in-progress improvements to micromatch that haven't been published quite yet (not sure whether that has an impact on your use case. 

Brace expansion is supported, but I haven't see an example yet with the leading comma like you showed. If you're having trouble with it in chokidar I suggest you open an issue there describing your pattern and the sorts of paths you expect it to match that aren't.

es128:
Btw another option for expressing what you want to watch with more standard syntax would be `['build/*.js', 'build/test/*.js']`

ohjames:
Thanks for the tip! I'm aware of that, I just saw the `{,optional-directory/}` pattern used in the angular-fullstack-generator and thought it was neat.

es128:
I just tried your pattern, and it seems to already work. Did you try it with chokidar before opening this issue?

```js
~/Code/anymatch (1.2.1)
λ node
> am = require('./')
[Function]
> am('build/{,test/}*.js', 'build/foo.js')
true
> am('build/{,test/}*.js', 'build/test/foo.js')
true
> am('build/{,test/}*.js', 'build/otherdir/foo.js')
false
```

ohjames:
@es128 Thanks for being so thorough with your help, yes I tested it with chokidar but not `1.0.1`, there's also of course the possibility I made a stupid mistake, I'll update this issue with my results in case you're interested, thanks again for your time.

es128:
Yes, if you can demonstrate a problem I'd like to hear about it.

ohjames:
@es128 here you go:

file: choki.js
```javascript
var chokidar = require('chokidar')
var watcher = chokidar.watch('src/{plugin/,}*.js')
[...]

Issue #614, opened 2017-06-13
Glob not working for multi-level paths within braces


bustbr:
Using the latest version (1.7.0) there's a bug in the [`filterDir`](https://github.com/paulmillr/chokidar/blob/1.7.0/index.js#L434) function of [`FSWatcher._getWatchHelpers`](https://github.com/paulmillr/chokidar/blob/1.7.0/index.js#L383).

Using the watcher like this:
```js
chokidar.watch('./{foo/bar,qoo}/**/*.js', opts)
```
will call [`dirParts = getDirParts(path)`](https://github.com/paulmillr/chokidar/blob/1.7.0/index.js#L430) with `path = '{foo/bar,qoo}/**/*.js'` which leads to `dirParts` being set to `['{foo', 'bar,qoo}', '**']`, which will never match any of the targeted files (e.g. `./foo/bar/test.js` or `./qoo/test.js`).

jonschlinkert:
To avoid splitting inside brackets or quoted strings, maybe we should consider using [split-string](https://github.com/jonschlinkert/split-string), with `options.brackets` set to `true`.

PR #622, opened 2017-07-17
fix for handling braces in path

MhMadHamster:


MhMadHamster:
should fix #287 and #614

es128:
Looks like it's causing a few tests to consistently fail.

Did you try a split-string based approach as suggested in #614?

MhMadHamster:
I took braces because of discussion in #287, I honestly did not test it on Linux, since my main windows, will try to fix tests asap.

MhMadHamster:
closing since i dont have macos so finding the problem causing the test fall would take some time for sure

MhMadHamster:
https://travis-ci.org/paulmillr/chokidar/jobs/260414604 this one looks really weird because in last build it was failing on braces test and i changed **only** braces test, @es128 can you maybe restard this particular job please, if you have access, will much appretiate it.

es128:
@MhMadHamster thanks this is really good work. I got the one job in Travis passing after a few retries - this isn't unusual, the CI envs are fickle, especially for OS X.

There is one last problem to solve - the "check ignore after stating" test is failing on Windows with this patch.

https://ci.appveyor.com/project/es128/chokidar/build/387

MhMadHamster:
@es128 tbh this test is failing for me even on the current release of chokidar, using windows 10 node 6.9.1 npm 3.10.8 😕 will look into that issue.

es128:
@MhMadHamster oh yeah? It wasn't failing on appveyor before this patch. I'm not sure what the issue is. Sporadic failures are just noise, but when the same test fails on every version in both modes it's indicative of something needing to be adjusted.

MhMadHamster:
@es128 yeah, I remember that it wasn't failing before, at least on in my environment.

es128:
Thanks for checking into it! ❤️

MhMadHamster:
Well, I tracked down problem down to anymatch tho I'm not sure what's the better way to resolve it, the problem with windows is: in the failing test we have variable testDir which is getting compared in ignore function to the path of watched dirs/files, the testDir itself is created using default nodejs path module, so the path inside using backslashes, tho when the function in ignored, getting called by anymatch, it's been called 2 times once with path "as is" and once with normalized path and I'm not sure what was idea behind that, therefore when we're calling the first time with value "as is", we're correctly getting false, but the second time we're calling with normalized path i.e. with forward slashes and getting true when we're expecting false.
The easy way to fix this is to use normalized path for test itself, tho as i said i'm not sure what was the idea behind calling once with path as is and once with normalized path, ideally, I think, path should be normalized in every case where it's being used, before doing anything with it.

es128:
So if you use `upath.normalizeSafe(path)` inside the `options.ignored` function in that test it'll fix it? I'd be fine with that change.

The path doesn't exactly change deliberately between ignore checks, it's just that the first time it's based on user input and the second time it's based on `readdirp` output and gets normalized. Separate work is going to have to be done to update glob-parent to latest and deal with the inconsistencies that result out of the windows paths. It doesn't need to be a barrier to landing this PR.

es128:
Although we do have to question why it worked before and why this patch had an impact.

MhMadHamster:
> So if you use upath.normalizeSafe(path) inside the options.ignored function in that test it'll fix it? I'd be fine with that change.

yes both for path passed in function and testDir

> Although we do have to question why it worked before and why this patch had an impact.

In fact this patch do not have impact on this test, i believe breaking changes can be found [here](https://github.com/es128/anymatch/commit/3c19a8b16f18431cfd501cd9e870a4dca007e6b5#diff-168726dbe96b3ce427e7fedce31bb0bcR34), since then check for normalized path was added.
So if you'll run chokidar tests with anymatch 1.3.0 it'll run just fine but if you update anymatch to 1.3.2 it'll fail (on windows only)

es128:
ooook I see, thanks for clarifying. Sorry for putting you through all this, and thank you for your diligence.

Let's just update the test to use the normalized path and land this PR.

MhMadHamster:
@es128 can you restart failing jobs please, thanks.

PR #631, opened 2017-08-17
bump is-glob

MhMadHamster:
bump is-glob version should fix some issues with parentheses, (!) [] etc in path, some tests for such cases need to be added probably

### DIFF
--- a/index.js
+++ b/index.js
@@ -8,6 +8,7 @@ var globParent = require('glob-parent');
 var isGlob = require('is-glob');
 var isAbsolute = require('path-is-absolute');
 var inherits = require('inherits');
+var braces = require('braces');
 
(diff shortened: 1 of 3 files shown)


MhMadHamster:
i think i did smth wrong while creating pr since the commits from the last one displaying as well :confused:

es128:
@MhMadHamster yeah it's because I used the "Squash and merge" button for your last PR, so if you kept your same branch going it's going to have a diverged commit history.

But if I do the same thing with this one it probably won't end up mattering.

MhMadHamster:
i'm not certainly sure in the correctness of current "fix", this needs more investigation

sayp:
Макс, красава, весь хотелс за тебя кулочки держит!

Issue #627, opened 2017-08-01
Bump glob-parent?


phated:
Just wondering if you were willing to bump glob-parent.  I believe version 3.1.0 is currently available.

/cc @es128

es128:
Willing, yes. There's more to it than a simple version bump - iirc it causes windows backslash problems without further adjustments within chokidar and I haven't taken the time to sort it out.

phated:
@es128 oh yeah, I totally forgot about that.  I'll ping @jonschlinkert and @doowb to see if they can look at this when they have time.

doowb:
I'll take a look this weekend.

phated:
@doowb sorry, I need to ping this again because this (and some other glob related dependencies) are a big duplication problem in my dependency tree.

@es128 any chance we could major bump all the outdated glob dependencies (anymatch, glob-parent, is-glob) and major bump here?  Even if it's a breaking change for my glob-watcher, I can just force the major in my tree.

phated:
This was done in 2.0.0

PR #632, opened 2017-08-21
Update globbing deps

es128:
Opening in response to #627 to help evaluate what's left to be solved.

@phated

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,13 +44,13 @@
     "fsevents": "^1.0.0"
   },
   "dependencies": {
-    "anymatch": "^1.3.0",
+    "anymatch": "^1.3.2",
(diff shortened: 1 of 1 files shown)


es128:
Bringing is-glob to latest is the only one that causes issues, tests still pass on all platforms with latest glob-parent.

Likely that the failing tests are wrong, using `path.join()` with globs. Easy enough to fix in the tests themselves.

The potential for users out there relying on invalid (but presently functional) globs that use backslash separators probably necessitates a major bump. 

I think in the past I was hoping to automatically correct the invalid globs somehow and not bump major, but unless someone comes forward with a clear way to accomplish that, I'm not going to worry about it at this point.

Issue #546, opened 2016-10-26
TypeError: Cannot read property 'lastChange' of undefined


bisol:
Got this error twice, not sure how to reproduce. This does not seems stop the watcher, and the same code runs fine on other systems. Node v5.12.0 on Ubuntu 10.04 with Chokidar 1.6.1

The app is configured to monitor a root dir with many subdirs (max depth = 1), each with many files (about 5Tb of 50kb to 30Mb files).

The error occurs on index.js at line 309:
      if (now - this._pendingWrites[path].lastChange >= threshold) {

Chokidar is configured thus:

watcher = chokidar.watch(path, {
  ignoreInitial: true,
  followSymlinks: false,
  awaitWriteFinish: true,
  ignorePermissionErrors: true,
  depth: 0
})
watcher.on('addDir', function (path) {
  watcher.add(path)
})

es128:
Looks like a bug occurring when the same file gets an event multiple times within the awaitWriteFinish polling interval.

If you can reproduce it reliably, try applying this patch and repeating the actions that would cause the error. Let me know if it works to prevent it.

```diff
diff --git a/index.js b/index.js
index e23a643..4b3fa0d 100644
--- a/index.js
+++ b/index.js
@@ -295,8 +295,8 @@ FSWatcher.prototype._awaitWriteFinish = function(path, threshold, event, awfEmit

   var awaitWriteFinish = (function (prevStat) {
     fs.stat(fullPath, function(err, curStat) {
-      if (err) {
-        if (err.code !== 'ENOENT') awfEmit(err);
+      if (err || !(path in this._pendingWrites)) {
+        if (err && err.code !== 'ENOENT') awfEmit(err);
         return;
       }
```

bisol:
Thanks for the quick response.
This happened two times yesterday, and again just now (after running for about 4 hours).
This app runs out of memory and restarts itself (using recluster) every few hours, so that may be the cause

es128:
It still indicates a flaw in the code that my proposed patch should resolve, but it would be nice to get some validation on that if you have the ability to try it.

bisol:
Yes, I should have mentioned I update my app with your patch and will redeploy it tomorrow (it's getting late here, and I don't own that server)

es128:
Resolved by #644 and soon to be published in v2.0.1

PR #644, opened 2017-10-14
Fix uncaught exception "Cannot read property 'lastChange' of undefined".

dsagal:
Fixes issue #546. Includes a test case that reproduces the issue before the fix.

Please note that the bug is severe. It causes an uncaught exception, for which the only recommended solution is to restart node. It is also rare and hard to reproduce: it seems to only happen in environments where fs.stat() might take some time to return. But for some environments + workloads it gets triggered regularly.

The test case introduces an artificial delay in fs.stat() to make sure we can trigger the bug reliably.

The fix proposed by @es128 in the discussion of paulmillr#546 does work perfectly, and is the only code change in this pull request.

### DIFF
--- a/index.js
+++ b/index.js
@@ -301,8 +301,8 @@ FSWatcher.prototype._awaitWriteFinish = function(path, threshold, event, awfEmit
 
   var awaitWriteFinish = (function (prevStat) {
     fs.stat(fullPath, function(err, curStat) {
-      if (err) {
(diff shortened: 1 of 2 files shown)


dsagal:
Test failures are mainly for "unwatch" test on MacOS, which I don't think is related. I'll try to re-run again.

paulmillr:
That change broke tests ever since.

PR #663, opened 2018-01-02
“Drop” support for nodejs <v6. We still support it but unofficially.

paulmillr:
cc @es128 @phated 

I think it's a reasonable move because basically no one supports it nowadays. Users who want to use it could fall back to v1.


phated:
Gulp 4 supports it.  It's not as hard as people think (basically avoid sindre modules and request) 😛

paulmillr:
Okay then!

es128:
@paulmillr icymi: https://github.com/paulmillr/chokidar/pull/658#issuecomment-354471869

Basically, nothing has really changed and it's not like we're burdened by supporting issues that come up that only affect old versions - to the extent we're even doing much to provide support at this point.

It may make sense at some point for there to be a rewrite or major refactor that leverages new features in core like `recursive` and to stop allowing for old versions along with that.

PR #669, opened 2018-01-12
Fix windows ignore when using anymatch@2 & cwd option

### LINKED ISSUE #668: Issue with absolute path under windows 10
Hello,

I'm facing an issue since upgrade to 2.0.0 on windows 10.
call to function
```

chokidar.watch("D:\my-library\app/**/*.ts").on('all', (event, path) => {
      console.log(path);
  });
```
the console.log is never called when editing a file despite it was working on version 1.7.0.

I had a look in the code, and it seems that in index.js l.420:
globFilter(resolvedPath); is returning false when called with a ts files whereas it should return true.

Please note also that if I called chokidar.watch() with a relative path instead of the global one, it works.
```
chokidar.watch("D:\my-library\app/**/*.ts") **fail**
chokidar.watch("app/**/*.ts") **succeed**
```

remy:
Fixes #668

Chokidar@2 includes anymatch@2 which has a (potentially) breaking change
as described in the changelog: https://github.com/micromatch/anymatch#change-log

When the `cwd` is set Chokidar will join the ignore pattern to the cwd,
and if this isn't absolute, it causes `anymatch` to fail:

```js
// where cwd = c:\Users\remy\
anymatch('C:\\Users\\remy\\**\\node_modules\\**',
'C:\\Users\\remy\\node_modules\\project'); // === false
```

https://runkit.com/embed/rn05fobk294p

So this change adds the upath module (used by Chokidar during testing
and about 35K footprint) to normalize the path to bash-like slashes
which corrects the issue.

### DIFF
--- a/index.js
+++ b/index.js
@@ -9,7 +9,8 @@ var isGlob = require('is-glob');
 var isAbsolute = require('path-is-absolute');
 var inherits = require('inherits');
 var braces = require('braces');
-var normalizePath = require('normalize-path')
+var normalizePath = require('normalize-path');
+var upath = require('upath');
 
 var NodeFsHandler = require('./lib/nodefs-handler');
 var FsEventsHandler = require('./lib/fsevents-handler');
@@ -357,7 +358,7 @@ FSWatcher.prototype._isIgnored = function(path, stats) {
     if (cwd && ignored) {
       ignored = ignored.map(function (path) {
         if (typeof path !== 'string') return path;
-        return isAbsolute(path) ? path : sysPath.join(cwd, path);
+        return upath.normalize(isAbsolute(path) ? path : sysPath.join(cwd, path));
       });
     }
     var paths = arrify(ignored)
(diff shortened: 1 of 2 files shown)


es128:
@remy thank you! This looks good at a glance, I'll take a closer look later.

remy:
Any joy on this PR? I saw the travis builds seem to bail every _other_ build for some reason.

es128:
Yeah that's just the flaky filesystems on Travis' OS X VMs. Typically just an annoyance. And Travis was having general issues when I tried to look the other day. The appveyor tests passed, which is the main thing that counts in this case.

remy:
Cheers! What's the turnaround for publishing? (just asking so I know when to publish a bump to nodemon too - mine's just automated through travis atm)

remy:
Been checking every now and then, and news on when/if this will be pushed to npm? Thanks.

es128:
@remy sorry for the delay. Chokidar v2.0.1 with this fix is now published.

remy:
🙏 thank you

PR #679, opened 2018-02-08
Accept upath minor and security patches

danthegoodman:
It looks like upath was accidentally limited to only v1.0.0, since it was just added in the last release.

If you have a reason for limiting the version so specifically, that's fine. I won't be offended by a PR rejection :) I just wanted to bring it to your attention.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,6 +54,6 @@
     "normalize-path": "^2.1.1",
     "path-is-absolute": "^1.0.0",
     "readdirp": "^2.0.0",
-    "upath": "1.0.0"
+    "upath": "^1.0.0"
   }
 }


es128:
@remy did you have any particular reason for not applying a semver range?

remy:
I didn't want to make any assumptions about your dev practises, so I applied the current version that was being used in tests successfully. Other than that - no.

es128:
I don't think it's going to matter in this case, but I'll go ahead and merge without publishing right away. Next time there's a release for some other reason this will be part of it, assuming it stays benign.

danielweck:
This is actually a pretty big deal, could you please release an update to `chokidar` soon?

In my project: `rm -rf node_modules && rm package-lock.json && npm install` =>

```
npm ERR! code ENOTSUP
npm ERR! notsup Unsupported engine for upath@1.0.0: wanted: {"node":">=0.10 <=7"} (current: {"node":"9.5.0","npm":"5.6.0"})
npm ERR! notsup Not compatible with your version of node/npm: upath@1.0.0
npm ERR! notsup Not compatible with your version of node/npm: upath@1.0.0
npm ERR! notsup Required: {"node":">=0.10 <=7"}
npm ERR! notsup Actual:   {"npm":"5.6.0","node":"9.5.0"}
```

:(

danielweck:
PS: `chokidar` is actually a deep dependency in my project, but with a bit of digging I discovered that this affects quite a few dependent packages.

danielweck:
Note that commonly-used projects like `webpack-dev-server` are affected too:

`npm-remote-ls webpack-dev-server@2.11.1 | grep -i upath`
```
   │  ├─ upath@1.0.0
```

(`grep -B 187 upath` to show the tree ancestry leading to `chokidar`)

danielweck:
Update: apparently `--engine-strict` can default to true in some cases!
`mkdir ChokidarTest && cd ChokidarTest && npm init --yes && npm install chokidar --engine-strict=false`
=> okay.
By forcing `--engine-strict` to false, I should be able to get around this issue.

danielweck:
v2.0.2 is out, fixes the `npm install --engine-strict=true` issue. See:
https://github.com/paulmillr/chokidar/issues/681#issuecomment-365670322

PR #684, opened 2018-02-21
Update braces to 2.3.1

maxcbc:
# Update braces to 2.3.1
Braces has 2.3.0 has a ReDos vulnerability: https://snyk.io/vuln/npm:braces:20180219
This is fixed 2.3.1.


es128:
This is unnecessary. The fix is covered in the semver range, and chokidar is used in locally-run build tools where I'm not very concerned about users trying to ReDos themselves.

PR #687, opened 2018-02-24
Add callbacks to fs.close

BridgeAR:
This also fixes an issue that using `fd = 0` is not closing so far.

Adding callbacks is important due to a soon coming [change](https://github.com/nodejs/node/pull/18668) in Node.js.


BridgeAR:
Ping @paulmillr

paulmillr:
Thank you @coveralls

PR #700, opened 2018-03-31
[wip] Add watchman backend

### LINKED ISSUE #486: Add Watchman option to chokidar
Hi,

After looking around different repos, there doesn't seem to be any movement on adding a watchman option to chokidar as was discussed here:

https://github.com/facebook/react-native/issues/628

and here:

https://github.com/ember-cli/ember-cli/issues/4503

Since chokidar supports symlinking, and sane does not (https://github.com/amasad/sane/issues/42), it would be nice to have an option to use watchman and if not fall back to chokidar. Also chokidar seems to perform better than sane.

Thanks.

ishitatsuyuki:
This is an experimental backend for watchman mainly intended for being overriden by the user. watchman is a robust and resource efficient file-watching server, which can be used for sharing inotify watches to avoid creating too much file descriptors.

We probably want to allow the user to specify the workspace root so we don't watch many subdirs.

69/90 test passing currently; globbing and unwatching has issues. symlinks are not supported at all, so its tests are disabled. Also, CI is currently left broken because watchman needs manual installation ([example](https://github.com/researchgate/webpack-watchman-plugin/blob/796a712b51647fc236798fa6c8a70193ac741e95/.travis.yml)).

Close #486

### DIFF
--- a/index.js
+++ b/index.js
@@ -14,6 +14,7 @@ var upath = require('upath');
 
 var NodeFsHandler = require('./lib/nodefs-handler');
 var FsEventsHandler = require('./lib/fsevents-handler');
+var WatchmanHandler = require('./lib/watchman-handler');
 
 var arrify = function(value) {
   if (value == null) return [];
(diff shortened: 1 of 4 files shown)


paulmillr:
good shit

aleclarson:
[filespy](https://github.com/alloc/filespy) (an alternative to chokidar) has bug-free Watchman support.

PR #704, opened 2018-04-20
Update fsevents to fix prototype pollution attack

fenichelar:
https://nodesecurity.io/advisories/566

https://github.com/strongloop/fsevents/pull/201
https://github.com/mapbox/node-pre-gyp/pull/347

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "sinon-chai": "^2.6.0"
   },
   "optionalDependencies": {
-    "fsevents": "^1.1.2"
+    "fsevents": "^1.2.0"
   },
   "dependencies": {
     "anymatch": "^2.0.0",


fenichelar:
A 2.0.4 release with this would be great :)

es128:
This is already handled by the caret semver operator, making this PR unnecessary. Just do a fresh install.

fenichelar:
It is best practice to release a new patch version to address a vulnerability because it forces commands like `npm update -S` to grab the new release without requiring a fresh install. When you have many applications to manage, it is really annoying to do fresh installs all over the place. I understand that this is not actually a security risk, but I think all vulnerabilities should be treated the same. Is there a reason that you seem to be so against making a new releases?

es128:
My main concern is that I do not yet know the fallout of the node 0.12 breaking change, and forcing fsevents@1.2.0 here makes that a less reconcilable situation for any users affected.

Sorry to cause you annoyance, but you do already have the means to resolve it at your disposal.

PR #706, opened 2018-04-23
don't use broken fsevents release

bjfletcher:
Hi there,

We've had some broken builds today using Chokidar because of a broken `fsevents` release. This PR ensures that we're using at least https://github.com/strongloop/fsevents/releases/tag/v1.2.2

All best,
Ben

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "sinon-chai": "^2.6.0"
   },
   "optionalDependencies": {
-    "fsevents": "^1.1.2"
+    "fsevents": "^1.2.2"
   },
   "dependencies": {
     "anymatch": "^2.0.0",


es128:
Can you show the logs of the failure? Most likely it was just a bad timing thing while I was dealing with the fsevents release. I doubt this PR is necessary.

bjfletcher:
We do hundreds of builds daily. Microservices. We can of course clear
caches with the build containers but this PR will enforce those that have
the broken release to update. Why not?

es128:
I haven't seen yet how merging this and preparing a new chokidar release would do anything to help resolve your concern. I think the issue has already been resolved.

bjfletcher:
No worries. It's just a practice I use a lot, especially when we've got a variety of caching in various places – developer machines, CI machines, and so on. So if we have a broken library release, then we have two choices: 1. clear the cache (meaning the machine would have to redownload everything again), or 2. update the semver to update just this one thing (this PR).

If you really don't want this PR then feel free to close. We'll just clear the caches that have the broken release (it's extra effort for us, but we don't have any other option and also I don't want you to feel that we don't appreciate your effort maintaining these great libraries :))

Have a good day 😎

bjfletcher:
Here's the issue: 

image

bjfletcher:
.20 IIRC

es128:
Ok so this isn’t directly related to the publishing problems I had this morning, the timing was just somewhat coincidental. 

I’m only on mobile right now, but when I get a chance a little later I’ll look at merging this and publishing a patch release for chokidar

PR #708, opened 2018-04-27
Drop support for Node.js < 4

realityking:
Complimentary to strongloop/fsevents#212, I think it's time to drop Node.js < 4.

The chokidar dependency upath has already dropped support Node.js < 4, this is currently causing CI to fail on 0.10 and 0.12. See these logs for an example: https://travis-ci.org/paulmillr/chokidar/jobs/372092794#L819

Note: Currently CI fails for all Node.js versions - but there are more failures on 0.10 and 0.12 than on the newer versions.

Also fsevents has dropped support for 0.12 (with 0.10 still being supported).


realityking:
@paulmillr Any interest in merging this?

No matter what happens in strongloop/fsevents#212, chokidar is currently broken on 0.10 and 0.12 anyway because upath already requires Node 4. See Travis tests here: https://travis-ci.org/paulmillr/chokidar/jobs/393728441

Between this one and #709 chokidar would also lose a couple of dependencies which is always nice.

paulmillr:
@realityking upath is devdependency, we need to check if chokidar still runs on v0.12 w/o devdeps

realityking:
@paulmillr I don't think it's a dev dependency. It's listed as a dependency

https://github.com/paulmillr/chokidar/blob/196142754809f3bd345c9e65cfbe3c66e959b385/package.json#L45-L58

And it's loaded in index.js

https://github.com/paulmillr/chokidar/blob/6e112b99f8a0aa4d3cdfd160e4f283a7eadfc50c/index.js#L13

paulmillr:
@realityking the URL you pointed to is related to `nyc`, not upath. Why do you think upath won't work? https://travis-ci.org/paulmillr/chokidar/jobs/393728441

realityking:
@paulmillr You're right! I must've misread the build log.

paulmillr:
live in 3.x branch

PR #709, opened 2018-04-27
Use ES2015 class syntax for FSWatcher

realityking:
Since #708 is breaking backwards compatibility, this seems like a good moment to make another tiny breaking change by switching to ES2015 class syntax.

The breaking change is that the class methods are not enumberable. This is unlikely to break anything.

Advantages are somewhat cleaner code - especially pulling out the `_globIgnored` getter out of the constructor is nice - and dropping the dependency on `inherits`.

Note: This will have to be rebased after #708 is merged.


paulmillr:
#798

PR #717, opened 2018-05-17
update upath version to avoid install errors on node v10

Gaafar:
upath versions lower than 1.0.5 fail to install on node v10


kouhin:
@paulmillr We are trying to upgrade to Node.js 10 in our product. Is there any plan to ship this feature?

paulmillr:
2.0.4 is out

Issue #552, opened 2016-11-10
Some add events were missing when copying many files into watched folder


xsq007:
When copying many files, e.g. hundreds of files into the watched  folder, I found number of add events is not equal to the total file number, i.e. some add events were missing.
Following is the configuration of chokidar.
```
  var watcher = chokidar.watch(watchDir, {
      persistent: true,
      ignoreInitial: false,
      followSymlinks: false,
      ignorePermissionErrors: true,
      awaitWriteFinish: true
    });
watcher.on('add',  path => {
  console.log('add' , path)
})
```
BTW, if I exit the program and re-run it,  then  all add events are emitted.  Guess ignoreInitial=false option does the trick.

Wondering if there is a way to prevent lost "add" events  when files are copied into the folder at first.

My environment is Windows 10 x-64 and Node.js v6.6.0

birjolaxew:
Cannot reproduce on Ubuntu 14.04, Node v4.6.2.

---

Steps:
1. In another folder, create 1000 files using `for i in $(seq 1 1000); do echo $i > "file-$i.txt"; done`. Copy those.
2. Run the code below, starting chokidar
3. Paste files into watched folder

Code:
```javascript
require('./').watch('test/', {
  ignoreInitial: true,
  ignored: /node_modules|\.git/
}).on('all', function(event, path) {
  console.log(event, path,"("+addCount+")");
  ++addCount;
}).on('ready', function() {
  console.log('Ready');
}).on('raw', console.log.bind(console, 'RawEvent:'));

var addCount = 0;
```

Observed:
```
...
add test/file-995.txt (995)
add test/file-996.txt (996)
add test/file-997.txt (997)
add test/file-998.txt (998)
add test/file-999.txt (999)
```

xsq007:
Tried to copy "grunt" folder in a node_modules folder. 
There are totally 504 files in the folder on Windows OS.

I tried several times and got different results of add events, e.g. 501, 499, 496, but no luck to get 504.
Possibly it is related with nested folder structure.

birjolaxew:
**Can replicate** on Ubuntu 14.04, Node v4.6.2.

---

Steps:
1. Do `npm install grunt` in a folder, copy every file inside `node_modules/grunt/node_modules/dateformat` (probably also works with other complex folders)
2. Run the below code in the main folder, with an empty folder called `test` present
3. Paste copied files into `test`

Code: 
```javascript
var watchPath = "test/";
var found = (function(){
[...]

PR #723, opened 2018-05-27
Fix issue #552

srguiwiz:
First wrote a patch to fix issue #552 against release 2.0.3.  See commit e2b720c .

When running tests found out similar fix was done in pull request #690 with different logic.

This here pull request matches pre-existing logic of chokidar.

And apparently #690 as merged in 5e75390 is flawed because it debounces **all** calls to `read()` together, even for different `directory`.  May be worth noting https://lodash.com/docs/#debounce .  It becomes clear as merged in 5e75390 if watching multiple directories, e.g. nested directories, that implementation's behavior becomes a mess, and wrong.

For clarity I arranged this in a clean separation into two commits:  First 28a70c4 undoing #690.  Then e2b720c showing this fix by itself.

### DIFF
--- a/index.js
+++ b/index.js
@@ -273,13 +273,18 @@ FSWatcher.prototype._throttle = function(action, path, timeout) {
     this._throttled[action] = Object.create(null);
   }
   var throttled = this._throttled[action];
-  if (path in throttled) return false;
+  if (path in throttled) {
+    throttled[path].count++;
+    return false;
+  }
   function clear() {
+    var count = throttled[path] ? throttled[path].count : 0;
     delete throttled[path];
     clearTimeout(timeoutObject);
+    return count;
   }
   var timeoutObject = setTimeout(clear, timeout);
-  throttled[path] = {timeoutObject: timeoutObject, clear: clear};
+  throttled[path] = {timeoutObject: timeoutObject, clear: clear, count: 0};
   return throttled[path];
 };
 
--- a/lib/nodefs-handler.js
+++ b/lib/nodefs-handler.js
@@ -4,7 +4,6 @@ var fs = require('fs');
 var sysPath = require('path');
 var readdirp = require('readdirp');
 var isBinaryPath = require('is-binary-path');
-var debounce = require('lodash.debounce');
 
 // fs.watch helpers
 
@@ -342,11 +341,14 @@ function(dir, stats, initialAdd, depth, target, wh, callback) {
   parentDir.add(sysPath.basename(dir));
   this._getWatchedDir(dir);
 
-  var debouncedRead;
-
   var read = function(directory, initialAdd, done) {
     // Normalize the directory name on Windows
     directory = sysPath.join(directory, '');
+
+    if (!wh.hasGlob) {
+      var throttler = this._throttle('readdir', directory, 1000);
+      if (!throttler) return;
+    }
 
     var previous = this._getWatchedDir(wh.path);
     var current = [];
@@ -378,10 +380,8 @@ function(dir, stats, initialAdd, depth, target, wh, callback) {
         this._addToNodeFs(path, initialAdd, wh, depth + 1);
       }
     }.bind(this)).on('end', function() {
+      if (throttler) throttler.clear();
       if (done) done();
-
-      // Run any pending reads that may be queued
-      debouncedRead.flush();
 
       // Files that absent in current directory snapshot
       // but present in previous emit `remove` event
@@ -401,13 +401,6 @@ function(dir, stats, initialAdd, depth, target, wh, callback) {
     }.bind(this)).on('error', this._handleError.bind(this));
   }.bind(this);
 
-  // Create a debounced version of read
-  debouncedRead = debounce(read, 1000, {
-    leading: true,
-    trailing: true,
-    maxWait: 1000
-  });
-
   var closer;
 
   if (this.options.depth == null || depth <= this.options.depth) {
@@ -415,21 +408,13 @@ function(dir, stats, initialAdd, depth, target, wh, callback) {
     closer = this._watchWithNodeFs(dir, function(dirPath, stats) {
       // if current directory is removed, do nothing
       if (stats && stats.mtime.getTime() === 0) return;
-      debouncedRead(dirPath, false);
+
+      read(dirPath, false);
     });
   } else {
     callback();
   }
-
-  // Close function that calls fs closer and cancels any pending debounced reads
-  return function () {
-    if (closer) {
-      closer();
-    }
-
-    // Cancel any pending reads that may be queued
-    debouncedRead.cancel();
-  };
+  return closer;
 };
 
 // Private method: Handle added file, directory, or glob pattern.
--- a/lib/nodefs-handler.js
+++ b/lib/nodefs-handler.js
@@ -380,7 +380,7 @@ function(dir, stats, initialAdd, depth, target, wh, callback) {
         this._addToNodeFs(path, initialAdd, wh, depth + 1);
       }
     }.bind(this)).on('end', function() {
-      if (throttler) throttler.clear();
+      var wasThrottled = throttler ? throttler.clear() : false;
       if (done) done();
 
       // Files that absent in current directory snapshot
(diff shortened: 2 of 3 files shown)


srguiwiz:
Inviting review by @es128 , @bmathews , @paulmillr , to avoid a defect in a recent merge to make it out in 2.0.4, that would be sad.

srguiwiz:
Here is an outline of what a test case should do, to prove correctness, and to prevent similar regressions in the future:

Create directories
a
a/b
a/b/c

Watch directory a, which also watches a/b and a/b/c.  Notably those subdirectories would get separate invocations of `_handleDir`.

Maybe wait a second.

Have code write a short text into files
a/a1.txt
a/a2.txt
a/a3.txt
a/a4.txt
a/b/b1.txt
a/b/b2.txt
a/b/b3.txt
a/b/b4.txt
a/b/c/c1.txt
a/b/c/c2.txt
a/b/c/c3.txt
a/b/c/c4.txt

It is important that code writing is in a fast loop.  Hence I suggest creating the directories first.  That is how I could reproduce the problem of #552 even in a single directory.  For the regression of #690 it takes multiple directories, three is better than two, but probably they don't have to be nested, as long as all are watched.

Could the file writing loop be too fast?  There shouldn't be a too fast or too slow for passing the test.  There is only a too fast or too slow to fail the test with bad watching code.  Really slow writing, one file every two seconds, will not fail.  Really really fast writing might pass in released version 2.0.3, and fail with pull request #690.  But instead of having a mental exercise, maybe someone should code it, run it, and report it.

The goal is to have a test case that should pass if chokidar is healthy.

A resulting test is expected to fail with released version 2.0.3, and with pull request #690.  It should pass with this pull request #723.

For now I challenge others to make this proof of correctness part of `test.js`.

es128:
Seems to make sense to me. Wish we could get more reports from people who experience this issue and who could report back how this patch works out when they try it.

srguiwiz:
Here are two attempts to write tests.  They pass always for me.  They pass too much to prove the fix.  Has to do with timing.  It appears they run too slow to fail.  Why do they take 3.5 seconds to run each?

As mentioned above, in this case something passing doesn't necessarily mean code is correct in all situations.  We know the defects are real.

It actually makes sense the first attempt below might be too fast [sic], the plain three writeFile in a row.  The second block below with callback to start next writeFile is closest to when we see defect in our actual app.  I would expect it to fail with release 2.0.3, and the nested directory test with #690.

One mystery is why do these tests take 3.5 seconds to run.  Rather than continue to investigate alone, below I post what I have.

Advice, @es128 , @paulmillr ?

First attempt:
```
    it('should emit three `add` events when three files were added in one directory', function(done) {
      var spy = sinon.spy();
      var test1Path = getFixturePath('add1.txt');
      var test2Path = getFixturePath('add2.txt');
      var test3Path = getFixturePath('add3.txt');
      watcher.on('add', spy).on('ready', w(function() {
        fs.writeFile(test1Path, Date.now(), simpleCb);
        fs.writeFile(test2Path, Date.now(), simpleCb);
        fs.writeFile(test3Path, Date.now(), simpleCb);
        waitFor([spy, 3], function() {
          spy.should.have.been.calledThrice;
          spy.should.have.been.calledWith(test1Path);
          spy.should.have.been.calledWith(test2Path);
          spy.should.have.been.calledWith(test3Path);
          done();
        });
      }));
    });
    it('should emit three `add` events when three files were added in three directories', function(done) {
      var spy = sinon.spy();
      var test1Path = getFixturePath('add1.txt');
      var testDir2Path = getFixturePath('b');
      var test2Path = getFixturePath('b/add2.txt');
      var testDir3Path = getFixturePath('c');
      var test3Path = getFixturePath('c/add3.txt');
      watcher.on('add', spy).on('ready', w(function() {
        fs.mkdirSync(testDir2Path, 0x1ed);
        fs.mkdirSync(testDir3Path, 0x1ed);
        fs.writeFile(test1Path, Date.now(), simpleCb);
        fs.writeFile(test2Path, Date.now(), simpleCb);
        fs.writeFile(test3Path, Date.now(), simpleCb);
        waitFor([spy, 3], function() {
          spy.should.have.been.calledThrice;
          spy.should.have.been.calledWith(test1Path);
          spy.should.have.been.calledWith(test2Path);
          spy.should.have.been.calledWith(test3Path);
          done();
        });
      }));
    });
```

Trying different timing, second attempt:
```
    it('should emit three `add` events when three files were added in one directory', function(done) {
      var spy = sinon.spy();
      var test1Path = getFixturePath('add1.txt');
      var test2Path = getFixturePath('add2.txt');
      var test3Path = getFixturePath('add3.txt');
      watcher.on('add', spy).on('ready', w(function() {
        fs.writeFile(test1Path, Date.now(), function() {
          fs.writeFile(test2Path, Date.now(), function() {
            fs.writeFile(test3Path, Date.now(), simpleCb);
          });
        });
        waitFor([spy, 3], function() {
          spy.should.have.been.calledThrice;
          spy.should.have.been.calledWith(test1Path);
          spy.should.have.been.calledWith(test2Path);
          spy.should.have.been.calledWith(test3Path);
          done();
        });
      }));
    });
    it('should emit three `add` events when three files were added in three directories', function(done) {
      var spy = sinon.spy();
      var test1Path = getFixturePath('add1.txt');
      var testDir2Path = getFixturePath('b');
      var test2Path = getFixturePath('b/add2.txt');
      var testDir3Path = getFixturePath('c');
      var test3Path = getFixturePath('c/add3.txt');
      watcher.on('add', spy).on('ready', w(function() {
        fs.mkdirSync(testDir2Path, 0x1ed);
        fs.mkdirSync(testDir3Path, 0x1ed);
        fs.writeFile(test1Path, Date.now(), function() {
          fs.writeFile(test2Path, Date.now(), function() {
            fs.writeFile(test3Path, Date.now(), simpleCb);
          });
        });
        waitFor([spy, 3], function() {
          spy.should.have.been.calledThrice;
          spy.should.have.been.calledWith(test1Path);
          spy.should.have.been.calledWith(test2Path);
          spy.should.have.been.calledWith(test3Path);
          done();
        });
      }));
    });
```

srguiwiz:
Have added test cases to pull request.

As stated above, if these test cases pass that doesn't prove complete correctness.  But if they fail, that is a failure.

On the machine I tested on these test cases fail with released chokidar 2.0.3, and also fail (though differently) with pull request #690.  All failures observed are as expected.  Note they may fail differently for non-polling versus polling.

These test cases pass with this pull request #723.

It took some time to successfully construct these test cases to show failure.  This matter is timing sensitive.  Therefore I suggest not to change these test cases gratuitously.

srguiwiz:
@es128 , @paulmillr , as test cases have been added, also described in this thread, can this pull request now be merged?  Most people use from npm, some folk may use HEAD, but probably no one will use a pull request, so it probably won't get more burn-in unless merged.

srguiwiz:
Sorry to back.  Improved test cases.  Still true that PR #723 passes, and released 2.0.3 fails.

Interesting new insight.  PR #723 is faster than PR #690, but both pass the tests.  After instrumenting PR #690 with these test cases, I observed debounce invokes separately for separate directories, hence passes the test.  An apparent reason is a separate instance of debounce per each invocation of _handleDir.  Hence, my earlier theory that PR #690 would break is no longer sustained.

While PR #690 passes the tests, PR #723 is more like the pre-existing chokidar, responds more quickly and reports closer to in order, because it more immediately re-runs `read()` if there was a rapidly following event.

Looks like your code was working, @bmathews .

srguiwiz:
This pull request #723 with **324ms** looks more responsive
```
✓ should emit nine `add` events when nine files were added in one directory (324ms)
✓ should emit thirtythree `add` events when thirtythree files were added in nine directories (836ms)
```
versus pull request #690 with **1172ms** tested with same tests in pull request #725 :
```
✓ should emit nine `add` events when nine files were added in one directory (1172ms)
✓ should emit thirtythree `add` events when thirtythree files were added in nine directories (1537ms)
```
One can make the point it is wrong to have an absolute time value in a non-polling algorithm.  Hence this pull request, not using an absolute time value, but using logic, is more responsive, without additional CPU load.

es128:
I'm fine with this whole PR. Thank you for the diligent work @srguiwiz.

Anyone else want to chime in?

bmathews:
Looks good to me!

PR #725, opened 2018-06-03
Add regression tests for adding several files

srguiwiz:
Only the tests in are this PR.  Can be used with existing master branch.

Also can be used if changing implementation with another pull request.


srguiwiz:
Closing because no longer needed, as identical tests since have been merged with PR #723.

Issue #730, opened 2018-06-11
chokidar causes app crash if removing nested directories in Node 10 in Windows


srguiwiz:
Using release 2.0.3 here is reproducible code.
Proof is at the end, crashes before getting to the end.
Occurs in Windows only.
Also proves probably should not blame fs.watch.
```
//

// this is about removing nested directories in Windows when chokidar is watching
// see https://github.com/paulmillr/chokidar/issues/730
//
// the problem is if after
/*
removing now
event, path  unlinkDir test-chokidar\nested\nested2
*/
//it exits the application in a crash, showing
/*
  - node  test-fs-watch.js[8916]: src\fs_event_wrap.cc:90: Assertion `wrap != nullptr' failed.
 1: node::DecodeWrite
 2: node::DecodeWrite
 3: node::Start
 4: v8::internal::interpreter::BytecodeDecoder::Decode
 5: v8::internal::RegExpImpl::Exec
 6: v8::internal::RegExpImpl::Exec
 7: v8::internal::Object::ToInt32
 8: v8::internal::Object::GetProperty
 9: v8::internal::NativesCollection<0>::GetScriptsSource
10: v8::internal::NativesCollection<0>::GetScriptsSource
11: ...
*/

//
// first prove fs.watch doesn't have the problem
//
setTimeout(() => {
  const fs = require("fs");

  const fse = require("fs-extra");

  if (process.platform !== "win32") {
    console.log("this code is meant to help find an issue with chokidar in Windows");
  } else {
    console.log("this code with fs.watch is expected to run fine, error EPERM is ok, doesn't crash");
  }

  fse.ensureDirSync("test-fs-watch/nested/nested2")

  const watchers = [];
  watchers.push(fs.watch("test-fs-watch"));
  watchers.push(fs.watch("test-fs-watch/nested"));
  watchers.push(fs.watch("test-fs-watch/nested/nested2"));

  for (let i = 0; i < watchers.length; i++) {
    let watcher = watchers[i];
    watcher.on("change", (eventType, filename) => console.log(`change ${1}: eventType, filename `, eventType, filename));
    watcher.on("error", (eventType, filename) => console.log(`error ${i}: eventType, filename `, eventType, filename));
  }

  setTimeout(() => {
    console.log("removing now");
    fse.remove("test-fs-watch/nested");
  }, 3000);

  setTimeout(() => {
    console.log("application was still running just fine");
    console.log("");
    //process.exit();
  }, 6000);
},
[...]

PR #731, opened 2018-06-12
Don't fs.FSWatcher.close() after a remove if Windows and Node >= 10

srguiwiz:
Could not tell whether this is the best possible fix for issue #730, but so far this apparently prevents crashes in some apps.

Submitting pull request to move towards being able to run apps without crashing.

Welcome discussion.

PR #732, opened 2018-06-12
Prevent watcher.close() crashing - fix Issue #730

srguiwiz:
Replace pull request #731 with a much cleaner fix, after reading new documentation at Node.js.

### DIFF
--- a/test.js
+++ b/test.js
@@ -7,6 +7,7 @@ var should = chai.should();
 var sinon = require('sinon');
 var rimraf = require('rimraf');
 var fs = require('graceful-fs');
+var fse = require('fs-extra');
 var sysPath = require('path');
 var upath = require("upath");
 var cp = require('child_process');
@@ -460,6 +461,26 @@ function runTests(baseopts) {
           expect(spy.args[0][1]).to.not.be.ok; // no stats
           rawSpy.should.have.been.called;
           if (!osXFsWatch010) spy.should.have.been.calledOnce;
+          done();
+        });
+      });
+    });
+    it('should emit two `unlinkDir` event when two nested directories were removed', function(done) {
+      var spy = sinon.spy();
+      var testDir = getFixturePath('subdir');
+      var testDir2 = getFixturePath('subdir/subdir2');
+      var testDir3 = getFixturePath('subdir/subdir2/subdir3');
+      fs.mkdirSync(testDir, 0x1ed);
+      fs.mkdirSync(testDir2, 0x1ed);
+      fs.mkdirSync(testDir3, 0x1ed);
+      watcher.on('unlinkDir', spy).on('ready', function() {
+        w(fse.remove.bind(fse, testDir2, simpleCb))(); // test removing in one
+        waitFor([spy], function() {
+          spy.should.have.been.calledWith(testDir2);
+          spy.should.have.been.calledWith(testDir3);
+          expect(spy.args[0][1]).to.not.be.ok; // no stats
+          rawSpy.should.have.been.called;
+          if (!osXFsWatch010) spy.should.have.been.calledTwice;
           done();
         });
       });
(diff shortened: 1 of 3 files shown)


test.js
```diff
@@ -464,6 +465,26 @@ function runTests(baseopts) {
...
+    it('should emit two `unlinkDir` event when two nested directories were removed', function(done) {
+      var spy = sinon.spy();
+      var testDir = getFixturePath('subdir');
+      var testDir2 = getFixturePath('subdir/subdir2');
+      var testDir3 = getFixturePath('subdir/subdir2/subdir3');
+      fs.mkdirSync(testDir, 0x1ed);
+      fs.mkdirSync(testDir2, 0x1ed);
+      fs.mkdirSync(testDir3, 0x1ed);
+      watcher.on('unlinkDir', spy).on('ready', function() {
+        w(fse.remove.bind(fse, testDir2, simpleCb))(); // test removing in one
```
es128: Can we do this without the added dependency?

It's just a convenience function, doing them both back to back with regular `fs` async calls will have the same effect.

es128:
Can you please drop a link in this thread to the documentation you relied upon?

test.js
srguiwiz: Wish it would work without `fs-extra`, but when I test with Node 10.1 I only get the failure with `fs-extra`, but didn't get with two async fs calls in a row.  Apparently something done differently in fs-extra, I didn't look.  Maybe it calls an OS command, I really didn't look under the hood, just don't know its internals.

So could we then say it is fs-extra's "fault"?  No, because same happens if deleting the directory that contains the other directory by user action in Windows Explorer.  So, it just happens fs-extra does whatever is more similar to when a user on the computer uses Windows Explorer to delete.

Reminder the issue we're fixing here is Windows only, apparently Node 10.0 to 10.4 only.  Tested to occur in 10.1 to 10.4, not in 8.11.2, and we now think won't occur in Node 10.4.1, not sure yet.

And the nasty part is the complete crash, exit from app, if a user deletes a directory that contains another directory, e.g. by using Windows Explorer.

Given the severe nastiness of that, use of fs-extra seems justifiable to have a good test case.  Only way known so far to reliably reproduce the defect.

A test that doesn't reproduce the would pass, but it wouldn't really test.

srguiwiz:
The documentation mentioned in this PR was discussed in withdrawn PR #731 while investigating.

Comment at https://github.com/nodejs/node/issues/21276 mentions a fix in Node.js master branch commit [cd8f06f](https://github.com/nodejs/node/commit/cd8f06f64f5fba32cf851de4d59c0e22f45b89c7) and apparently will be coming with Node 10.4.1 per commit [8be3c97](https://github.com/nodejs/node/commit/8be3c97d41a7afff712f041ed577516a3b9b914a).

Reading the commit at Node reveals

> The errored `fs.FSWatcher` object is no longer usable in the event handler.

test.js
es128: Did you try `rimraf`? Apparently that's all that's going on under `fse.remove` - unless they have some modified implementation of that standard dependency that we already use here.

srguiwiz: You are right, `rimraf` does it as well.  Committed test using `rimraf` instead of `fs-extra`.

srguiwiz:
The related fix at Node has moved to commit [c09bfd8](https://github.com/nodejs/node/commit/c09bfd81b7e9a6f8fc0909e2eca5c1b35152811e) which didn't make it into 10.4.1, hopefully on that branch [v10.x-staging](https://github.com/nodejs/node/commits/v10.x-staging) will go into 10.4.2.  Either way, documentation in that related fix at Node says explicitly, after an `error` or `close` event to no longer use a watcher.  The problem is on Windows when deleting a watched directory that contains another directory (a simple occurrence in real life) then an `EPERM` error causes that to happen in chokidar.  So, any app using chokidar in Windows with Node 10.0 to 10.4.something will crash when a user in Windows Explorer deletes a watched directory that contains another directory, and that is a **hard crash**, no `try catch` can prevent it.  This PR solves the problem.

PR #737, opened 2018-06-18
Add tests to narrowly demonstrate defect mentioned in PR #682

srguiwiz:
This isn't a fix yet, but more narrowly demonstrates scope of defect mentioned in PR #682.

The referenced broadly failing test could create an impression this software library had gotten worse.  In fact, it hadn't gotten worse.  These tests now spell out more narrowly what still works as always and continues to work, versus what is broken and apparently had been broken for some time.

### DIFF
--- a/test.js
+++ b/test.js
@@ -641,7 +641,51 @@ function runTests(baseopts) {
       var unlinkSpy = sinon.spy(function unlinkSpy(){});
       var addSpy = sinon.spy(function addSpy(){});
       var testPath = getFixturePath('unlink.txt');
-      watcher = chokidar.watch([testPath, testPath + '.does-not-exist'], options)
+      watcher = chokidar.watch([testPath], options)
+        .on('unlink', unlinkSpy)
+        .on('add', addSpy)
+        .on('ready', function() {
+          w(fs.unlink.bind(fs, testPath, simpleCb))();
+          waitFor([unlinkSpy], w(function() {
+            unlinkSpy.should.have.been.calledWith(testPath);
+            w(fs.writeFile.bind(fs, testPath, 're-added', simpleCb))();
+            waitFor([addSpy], function() {
+              addSpy.should.have.been.calledWith(testPath);
+              done();
+            });
+          }));
+        });
+    });
+    it('should detect unlink and re-add unfazed by watching a second file', function(done) {
+      options.ignoreInitial = true;
+      var unlinkSpy = sinon.spy(function unlinkSpy(){});
+      var addSpy = sinon.spy(function addSpy(){});
+      var testPath = getFixturePath('unlink.txt');
+      var siblingPath = getFixturePath('sibling.txt');
+      fs.writeFileSync(siblingPath, 'sibling');
+      watcher = chokidar.watch([testPath, siblingPath], options)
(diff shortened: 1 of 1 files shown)


es128:
This looks excellent, thank you. However, I'm not so sure about that last commit. We want `ignoreInitial` for these tests to ensure that events coming before the `ready` event do not interfere with the spies

test.js
```diff
@@ -636,12 +636,97 @@ function runTests(baseopts) {
           });
         });
     });
+    it('should detect unlink unfazed by in other directory watching a second file that does not exist', function(done) {
```
es128: little phrasing tweaks

`should detect unlink while watching a non-existent second file in another directory`

test.js
```diff
@@ -636,12 +636,97 @@ function runTests(baseopts) {
...
+            unlinkSpy.should.have.been.calledWith(testPath);
+            w(fs.writeFile.bind(fs, testPath, 're-added', simpleCb))();
+            waitFor([addSpy], function() {
+              addSpy.should.have.been.calledWith(testPath);
+              done();
+            });
+          }));
+        });
+    });
+    it('should detect unlink and re-add unfazed by watching a second file', function(done) {
```
es128: `should detect unlink and re-add while watching a second file`

test.js
```diff
@@ -636,12 +636,97 @@ function runTests(baseopts) {
...
+            unlinkSpy.should.have.been.calledWith(testPath);
+            w(fs.writeFile.bind(fs, testPath, 're-added', simpleCb))();
+            waitFor([addSpy], function() {
+              addSpy.should.have.been.calledWith(testPath);
+              done();
+            });
+          }));
+        });
+    });
+    it('should detect unlink and re-add unfazed by in other directory watching a second file that does not exist', function(done) {
```
es128: `should detect unlink and re-add while watching a non-existent second file in another directory`

test.js
```diff
@@ -636,12 +636,97 @@ function runTests(baseopts) {
...
+            unlinkSpy.should.have.been.calledWith(testPath);
+            w(fs.writeFile.bind(fs, testPath, 're-added', simpleCb))();
+            waitFor([addSpy], function() {
+              addSpy.should.have.been.calledWith(testPath);
+              done();
+            });
+          }));
+        });
+    });
+    it('should detect unlink and re-add unfazed by in same directory watching a second file that does not exist', function(done) {
```
es128: `should detect unlink and re-add while watching a non-existent second file in the same directory`

srguiwiz:
The very last test added originated from a thought experiment.  Admittedly, it didn't prove a suspicion.  Nevertheless, a correct implementation should pass that test too.  A good fix should pass this test without a problem, a bad fix would have to fear it.

srguiwiz:
A main objective here was to make clear there was no regression.  Seeing 2.0.4 released, that goal appears achieved.

There is "only" a newly discovered defect.

The additional test cases here I have added for sanity, as guidance for implementing a fix.

I have tried implementing a fix too, as a public service, but haven't figured one.

Watching individual files is not a current use case for our work, so this is not a priority for me, and I will not proceed trying to implement a fix, not at this time.

PR #741, opened 2018-06-28
chore(deps): updated dep for compatibility node10

DalderupMaurice:
Since the 1.1.0 release of upath the engine requirement `<=9` has been deleted. This 1.1.0 release is/was not covered with the caret version range. 

This PR updates the dependency to make babel 7 compatible with node version above 10 
(Babel 7 uses chokidar as a dependency and using babel with node 10 and above results in an error, which this PR should resolve)

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,6 +54,6 @@
     "normalize-path": "^2.1.1",
     "path-is-absolute": "^1.0.0",
     "readdirp": "^2.0.0",
-    "upath": "^1.0.5"
+    "upath": "^1.1.0"
   }
 }


es128:
> This 1.1.0 release is/was not covered with the caret version range

That isn't true, semver caret only stops at major version bumps. Tilde would be the case of not including minor version bumps.

If you're not seeing the update, you may be having issues with lockfiles.

DalderupMaurice:
Lawl - completely overlooked we have a nested package.json, containing an older version 🙄 Sorry!

Issue #745, opened 2018-07-14
Watch does not work for glob pattern '<dir>/**' if '<dir>' does not exist on fs


eGavr:
**Describe the bug**

Files are not being watched for glob `<dir>/**` if `<dir>` does not exist.

**Versions (please complete the following information):**
 - Chokidar version [2.0.4]
 - Node version [8.6.0]
 - OS version: [Ubuntu 16.04]

**To Reproduce**

```js
// index.js
const chokidar = require('chokidar');
const watcher = chokidar.watch(['dir/**']);

watcher.on('add', path => console.log(path));
```

Run this code and create `file.ext` in `dir`:

```bash
mkdir dir && touch dir/file.ext
```

You will see nothing in console.

paulmillr:
Duplicate of latter-created #872

PR #746, opened 2018-07-14
fix: watch does not work for glob pattern '<dir>/**' if '<dir>' does not exist on fs

eGavr:
Resolves https://github.com/paulmillr/chokidar/issues/745

### DIFF
--- a/index.js
+++ b/index.js
@@ -390,7 +390,8 @@ FSWatcher.prototype._isIgnored = function(path, stats) {
 var replacerRe = /^\.[\/\\]/;
 FSWatcher.prototype._getWatchHelpers = function(path, depth) {
   path = path.replace(replacerRe, '');
-  var watchPath = depth || this.options.disableGlobbing || !isGlob(path) ? path : globParent(path);
+  var getExistingPath = function(path) { return fs.existsSync(path) ? path : getExistingPath(sysPath.dirname(path)); };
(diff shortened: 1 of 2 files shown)


srguiwiz:
@eGavr , thank you for taking initiative and fixing.

Testing is necessary.  Testing this project for now is a little bit odd, because the test suite has cases in it that don't pass, for yet to be fixed defects.

In order to ensure not to break existing functionality with a next release, please:

- Run test suite with your fix not in (previous commit), note what fails.
- Run test suite with your fix.  Compare what fails then.

That protects quality and sanity.

To be more perfect, you could also add a test case in test.js that fails before your fix and then passes with your fix.

eGavr:
@srguiwiz 

I've created a testing pull request https://github.com/paulmillr/chokidar/pull/749 where I've just changed  README and you can see in it that there are 7 broken tests.

In my PR there were 7 broken tests too. 
Then I've wrote a test on my changed, push it and I see that there are 6 broken tests for now :)

I think that some the test are not stable and fail randomly.

Can you review my chnages, please? It seems that they do not break the current functionality and fix the problem which I've described in the issue.

eGavr:
For now there are 9 broken tests :(

The previous commit which is different from the latest one only in test name has 7 broken tests.

eGavr:
ping :)

PR #748, opened 2018-07-14
fix: watch does not work for glob pattern which intersects with non-glob part of another glob pattern

### LINKED ISSUE #747: Watch does not work for glob pattern which  intersects with non-glob part of another glob pattern
**Describe the bug**

Files are not being watched for glob `<dir>/**` if `<dir>` does not exist.

 - Chokidar version [2.0.4]
 - Node version [8.6.0]
 - OS version: [Ubuntu 16.04]

**To Reproduce**

```js
// index.js
const watcher = chokidar.watch([
    'some-*/*/*.js',
    'some-dir/*/*.css'
]);

watcher.on('add', path => console.log(path));
```

Create files `some-dir/dir/file.css`, `some-dir/dir/file.js` and run the script.

You will see nothing in console:

```
some-dir/dir/file.css
```

No any information about `some-dir/dir/file.js`.

If then you add, for example, `some-dir/dir/another-file.js` nothing logs to console too.

eGavr:
Resolves #747

### DIFF
--- a/lib/nodefs-handler.js
+++ b/lib/nodefs-handler.js
@@ -374,7 +374,7 @@ function(dir, stats, initialAdd, depth, target, wh, callback) {
       // Files that present in current directory snapshot
       // but absent in previous are added to watch list and
       // emit `add` event.
-      if (item === target || !target && !previous.has(item)) {
+      if (item === target || !target && (!previous.has(item) || entry.stat.isDirectory())) {
         this._readyCount++;
 
         // ensure relativeness of path is preserved in case of watcher reuse
--- a/test.js
+++ b/test.js
@@ -1007,6 +1007,27 @@ function runTests(baseopts) {
           });
         });
     });
+    it('should correctly handle intersecting of glob patterns with non-glob parts of other glob patterns', (done) => {
+      var spy = sinon.spy();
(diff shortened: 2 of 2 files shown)


srguiwiz:
@eGavr , thank you for taking initiative and fixing.

Testing is necessary.  Testing this project for now is a little bit odd, because the test suite has cases in it that don't pass, for yet to be fixed defects.

In order to ensure not to break existing functionality with a next release, please:

- Run test suite with your fix not in (previous commit), note what fails.
- Run test suite with your fix.  Compare what fails then.

That protects quality and sanity.

To be more perfect, you could also add a test case in test.js that fails before your fix and then passes with your fix.

eGavr:
@srguiwiz 

I've created a testing pull request https://github.com/paulmillr/chokidar/pull/749 where I've just changed  README and you can see in it that there are 7 broken tests.

In my PR there were 7 broken tests too. 
Then I've wrote a test on my changed, push it and I see that there are 6 broken tests for now :)

I think that some the test are not stable and fail randomly.

Can you review my chnages, please? It seems that they do not break the current functionality and fix the problem which I've described in the issue.

eGavr:
ping :)

eGavr:
I've found some problem
Trying to fix it

Issue #751, opened 2018-07-18
Change event is not triggered when watching a symlinked directory with a dot in the dirname


joeshehata:
**Describe the bug**
I have a directory in node_modules that is a symlink to a different directory. The directory has a dot in the name (eg. `node_modules/@some-scope/contains.dot`. When I watch the symlinked directory, I get the `addDir` and `add` events, but no change events fire.

I haven't done enough digging into the codebase here, but I did run into this line: https://github.com/paulmillr/chokidar/blob/master/lib/fsevents-handler.js#L41

It looks like the intention of this line is to check if the path is a file or a directory. `sysPath.extname` changes the `watchPath` to `node_modules/@some-scope` to  When I change this line to use `fs.statSync(path).isDirectory()` (with some more error handling), everything seems to work

**Versions (please complete the following information):**
 - Chokidar version 2.0.4
 - Node version 8.11.2
 - OS version: macOS High Sierra 10.13.5

**To Reproduce**
I created a pull request with a failing test case in it: https://github.com/paulmillr/chokidar/pull/752

**Expected behavior**
Symlinked directories should fire change events as long as it is a valid directory

PR #752, opened 2018-07-18
Adds test for symlinks with dots in the dirname

joeshehata:
This is the test case added for the following issue: https://github.com/paulmillr/chokidar/issues/751

### DIFF
--- a/lib/fsevents-handler.js
+++ b/lib/fsevents-handler.js
@@ -28,6 +28,11 @@ function createFSEventsInstance(path, callback) {
   return (new fsevents(path)).on('fsevent', callback).start();
 }
 
+// safely determines if path is a directory
+function isDirectory (path) {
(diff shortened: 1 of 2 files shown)


joeshehata:
I went ahead and added a solution to this as well. I had some failing tests before adding my fix. All of the tests pass, except the ones that were already failing.

lib/fsevents-handler.js
```diff
@@ -28,6 +28,11 @@ function createFSEventsInstance(path, callback) {
   return (new fsevents(path)).on('fsevent', callback).start();
 }
 
+// safely determines if path is a directory
+function isDirectory (path) {
+  return fs.existsSync(path) && fs.statSync(path).isDirectory()
```
paulmillr: We don't want to use sync IO.

PR #762, opened 2018-08-29
ignore accessTime updates caused by read operations (fix #750)

### LINKED ISSUE #750: Chokidar will trigger for unmodified files on Windows when LastAccessTime is changed
Given the following in Windows:

    var watcher = chokidar.watch(dir, {
        ignored: ignored,
        persistent: true,
        usePolling: config.options.legacyWatch || false,
        interval: config.options.pollingInterval,
    });

When a file is opened but not modified, Chokidar will trigger a change as `LastAccessTime` is modified. All other attributes remain unchanged. See https://github.com/remy/nodemon/issues/1208

 - Chokidar version: 170
 - Node version: v8.11.2
 - OS version: Microsoft Windows NT 10.0.17713.0

**To Reproduce**
Just open a file (changing `LastAccessTime` on a watched file. LMK if you want a demo app.

**Expected behavior**
No changes. Access Time is not considered a change. It is reading, not writing.

markablov:


mikemaccana:
Thanks @paulmillr! If you publish a release using this we can get `nodemon` to use it

PR #791, opened 2019-02-08
A better workaround for atomic writes

### LINKED ISSUE #237: Only a single `change` event is fired for each file
Here's a very basic example:

``` javascript
require('chokidar').watch(process.argv).on("change", function(path) {
  console.log(path);
});

require('http').createServer().listen(8000);
```

After I run this code and start changing watched files I see the events being fired. The problem is that only the first change for each file is reported!

I have this issue with versions 0.12.* and 1.0.0-rc*. Versions 0.11.* and below seem to work fine.

I'm using Arch Linux with local Btrfs and Ext4 filesystems (tested both). Kernel: 3.18.6-1-ARCH, node: 0.10.36.

With `usePolling` option everything works fine.

### LINKED ISSUE #742: Changes stopped being reported after the first change event
**Describe the bug**
Chokidar stops reporting a file's changes after the reporting first change event.

* Chokidar version 2.0.4
* Node version 10.6.0
* OS: Arch Linux

**To Reproduce**
Steps to reproduce the behavior.
1. Watch a **single file**
2. Modify it (the change would be reported)
3. Modify it again

From here no changes will be reported when the target is updated.

Example:
```
const chokidar = require("chokidar");
const watcher = chokidar.watch("test.txt");

watcher.on("change", filepath => {
    console.log("Update:", filepath);
});
```

**Expected behavior**
The watcher should report changes.

**Additional context**
* Renaming the file, then renaming it back avoids this.
* I would be very surprised if this is the expected behavior.

mitar:
This fixes #742 and fixes #237.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,7 +132,7 @@ chokidar.watch('file', {
   },
 
   ignorePermissionErrors: false,
-  atomic: true // or a custom 'atomicity delay', in milliseconds (default 100)
+  atomic: true
 });
 
 ```
@@ -221,12 +221,8 @@ Use with caution.
 that don't have read permissions if possible. If watching fails due to `EPERM`
 or `EACCES` with this set to `true`, the errors will be suppressed silently.
 * `atomic` (default: `true` if `useFsEvents` and `usePolling` are `false`).
-Automatically filters out artifacts that occur when using editors that use
-"atomic writes" instead of writing directly to the source file. If a file is
-re-added within 100 ms of being deleted, Chokidar emits a `change` event
-rather than `unlink` then `add`. If the default of 100 ms does not work well
-for you, you can override it by setting `atomic` to a custom value, in
-milliseconds.
+Uses a workaround to handle file changes by editors that use
+"atomic writes" instead of writing directly to the source file.
 
 ### Methods & Events
 
--- a/index.js
+++ b/index.js
@@ -113,9 +113,8 @@ function FSWatcher(_opts) {
     opts.interval = parseInt(envInterval);
   }
 
-  // Editor atomic write normalization enabled by default with fs.watch
+  // Editor atomic write handling enabled by default with fs.watch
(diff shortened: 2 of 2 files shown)


SGrondin:
Posting here to confirm that this PR fixed our issues in Parcel.

paulmillr:
@SGrondin fixed issues in Parcel and broke in Gatsby...

cmcaine:
@paulmillr do you have any ideas for how to fix this without breaking other tools? It's unclear to me why gatsby failed...

I might even have the time to do a PR ;)

Refs:
https://github.com/gatsbyjs/gatsby/pull/12759
https://github.com/parcel-bundler/parcel/issues/221

In the meantime #742 and #237 should probably be re-opened, right?

wpyoga:
Hello. I noticed a problem with docusaurus a few days ago, and traced it back to this apparent problem in chokidar. This is the issue: facebook/docusaurus#5252 . I've set up a repo to reproduce the issue: [wpyoga/repro-docusaurus-5252](https://github.com/wpyoga/repro-docusaurus-5252), hopefully that can help to diagnose the problem. I've only tried it on my Linux laptop though.

The problem manifests itself in this way: when I run the docusaurus development server, and then make changes to its config file `docusaurus.config.js` using `vim`, only the first change will trigger a rebuild of the site. Further changes wouldn't. Also, other editors including `nano` and VS Code don't trigger this issue.

Based on my experiments, what I think happened was:
1. When `vim` saves a file, it will write the data to a temporary file, then remove the original file, and then replace the original file with the temporary file.
1. This `setTimeout()`'s callback runs before the delay value passed into `setTimeout()`: https://github.com/paulmillr/chokidar/blob/master/index.js#L579-L585 . This caused the file to be removed from the watch list, so further changes to the file does not trigger an event.
1. I was able to reproduce this in my repo.
1. Outside of chokidar, I could not make `setTimeout()` execute its callback before the delay specified.

PR #796, opened 2019-02-11
handle ENOSPC with try catch

souparno:


paulmillr:
I don't think we should handle ENOSPC. This error specifically means "free up your hard disk, or bad shit would happen". Chokidar is one of many, many apps that would crash. And chokidar doesn't even create new files on disk, so we are not participating in hard disk storage fill.

souparno:
@paulmillr ENOSPC is not always thrown cause of the low disk space, it is also thrown when the maximum number of system watchers exceeds the io limit set.
it is also encountered when trying to watch large enough file system.
https://github.com/facebook/jest/issues/3254#issuecomment-297214395

paulmillr:
Good point. I don't think we should let it fail silently in this case. Any other ideas?

PR #818, opened 2019-04-02
fix #806: Chokidar doesn't let process exit when watching globs with subfolders

### LINKED ISSUE #806: Chokidar doesn't let process exit when watching globs with subfolders
**Describe the bug**

When watching globs of files (e.g. `**`, `**/*.js` or any other glob that works with subfolders), chokidar's `watcher.close()` doesn't seem to clean up correctly and prevents node process from exiting.

 - Chokidar version: 2.1.2
 - Node version: v11.10.0 and v10.13.0
 - OS version: Ubuntu 18.04, Windows 10

**To Reproduce**
Just run the following code:
```
const watcher = chokidar.watch('**');
setTimeout(() => watcher.close(), 1000);
```

Note that changing the globe to a "flat" one (e.g. `*`, `*.js` or any other that _doesn't_ access subfolders) makes chokidar exit correctly.

**Expected behavior**
Chokidar should cleanup listeners correctly and allow node process to exit.

**Additional context**
Possibly related issues: #688, #645

qraynaud:
This is a simple change. I noticed that when some patterns needed to watch for different files in the same directory, a single watcher could add multiple listeners on the same directory.

They all need to be closed and not only one because if not the listeners count never gets to `0` and the underlying `fs.watch` watcher is never disposed.

Since I don't know well the overall codebase I don't know if it's normal that a single chokidar watcher does multiple calls to `setFsWatchListener` with the same `path` but it obviously does.

### DIFF
--- a/index.js
+++ b/index.js
@@ -580,7 +580,9 @@ FSWatcher.prototype._remove = function(directory, item) {
 
 FSWatcher.prototype._closePath = function(path) {
   if (!this._closers[path]) return;
-  this._closers[path]();
+  this._closers[path].forEach(function(closer) {
+    closer();
+  });
   delete this._closers[path];
   this._getWatchedDir(sysPath.dirname(path)).remove(sysPath.basename(path));
 }
@@ -699,7 +701,9 @@ FSWatcher.prototype.close = function() {
 
   this.closed = true;
   Object.keys(this._closers).forEach(function(watchPath) {
-    this._closers[watchPath]();
+    this._closers[watchPath].forEach(function(closer) {
+      closer();
+    });
     delete this._closers[watchPath];
   }, this);
   this._watched = Object.create(null);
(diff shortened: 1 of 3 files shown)


qraynaud:
I also looked into the 3.x codebase and this will need to be ported over if we want to keep this working. I did not write a test to ensure there is no regression on this later on, sorry...

qraynaud:
I don't understand why the tests are not green. Especially the nod v10 tests on Linux. This is my setup and it works very well on my computer! Any ideas?

paulmillr:
@qraynaud the tests are shitty, it's not your fault. they are passing on my mac also.

qraynaud:
Is it possible to make a 2.1.6 release so that projects that my dependencies that currently uses `"chokidar": "^2.0.0"` as their dependency pattern gets this fix?

That would help a lot!

Thanks!

qraynaud:
@paulmillr Any idea about if this is possible?

Issue #797, opened 2019-02-12
Chokidar 3.x


paulmillr:
- Support only node 8+; since recent fsevents would do the same
- Async / await instead of callbacks
- Less code, less dependencies

Globegitter:
Would it also be possible to get a/the fix in for https://github.com/paulmillr/chokidar/issues/696?

paulmillr:
We'll try; any help on the issue would be appreciated.

Globegitter:
As I said I could only reproduce it on macos but it works fine on linux. I also posted a fix: https://github.com/ecosia/chokidar/tree/symlink-fix-mac we have been using this now for months on the mac's of our developers and it seemingly is working fine. I am also happy to make a PR out of this.

paulmillr:
Out!

PR #822, opened 2019-04-17
Symlink fix mac

### LINKED ISSUE #696: Symlinks not followed
Symlinked files are not followed, even with followSymlinks explicitly set to true.

You may use the following code to replicate the problem:
```

var watchFiles = [
  "path/to/non-symlinked/file",
  "path/to/symlinked/file",
]
var watcher = chokidar.watch(watchFiles, {
  ignored: /(^|[\/\\])\../,
  persistent: true,
  followSymlinks: true
});

// Something to use when events are received.
var log = console.log.bind(console);

// Add event listeners.
watcher
  .on('add', path => log(`File ${path} has been added`))
  .on('change', path => log(`File ${path} has been changed`))
  .on('unlink', path => log(`File ${path} has been removed`));
```
When I make a change to the non-symlinked file, I get a console message that the file has been changed. When I make a change to the symlinked file, no event bubbling occurs and no console message appears. If I adjust the symlink itself, then event bubbling does occur.

This issue is significant to my team because we are using Bazel to build all of our projects. Bazel runs the development server in a sandbox where it creates symlinks to all of the files in our app.

My environment
| Software         | Version(s) |
| ---------------- | ---------- |
| Chokidar          | 2.0.2
| Node             | 9.7.1
| npm/Yarn         | 5.6.0
| Operating System | High Sierra

Globegitter:
As mentioned in #797 and #696 this the fork we are currently running on that fixes #696. Would be great to get this or a similar fix into upstream for 3.0

cc @paulmillr

### DIFF
--- a/lib/fsevents-handler.js
+++ b/lib/fsevents-handler.js
@@ -285,16 +285,7 @@ function(linkPath, fullPath, transform, curDepth) {
 
     // add the linkTarget for watching with a wrapper for transform
     // that causes emitted paths to incorporate the link's path
-    this._addToFsEvents(linkTarget || linkPath, function(path) {
-      var dotSlash = '.' + sysPath.sep;
-      var aliasedPath = linkPath;
-      if (linkTarget && linkTarget !== dotSlash) {
-        aliasedPath = path.replace(linkTarget, linkPath);
-      } else if (path !== dotSlash) {
-        aliasedPath = sysPath.join(linkPath, path);
-      }
-      return transform(aliasedPath);
-    }, false, curDepth);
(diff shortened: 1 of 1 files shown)


paulmillr:
need to resolve merge conflicts for this.

paulmillr:
Also, can you add actual tests for this feature? That would be broken in master and fixed in your branch.

paulmillr:
I don't think that fixes it for real.

`ln -s from to`

Right now chokidar is checking whether `ignoredFiles` contain `from`. You're changing it to check for `to`. That isn't right.

PR #843, opened 2019-05-04
Rewrite with promises

JSMonk:


index.js
```diff
@@ -320,6 +325,18 @@ constructor(_opts) {
 
 // Public methods
 // --------------
+emit(...args) {
+  super.emit(...args);
```
paulmillr: Why?

index.js
```diff
@@ -876,5 +887,10 @@ exports.FSWatcher = FSWatcher;
  * @param {Object=} options chokidar opts
  * @returns an instance of FSWatcher for chaining.
  */
-const watch = (paths, options) => new FSWatcher(options).add(paths);
+const watch = (paths, options) => {
+  const watcher = new FSWatcher(options);
+  watcher.add(watcher._normalizePaths(paths));
```
paulmillr: Why are we doing this here and not in `watcher.add`?

Issue #469, opened 2016-04-24
NO `ready` event if directory contains a symlink on Windows


CxRes:
Like the title says:

On Windows 10, Chokidar 1.4.3, nodejs 5.7.1 - 5.11.0

CxRes:
A little evolution on this: I do not get `ready` only when the symlink directory does not exist. I am wondering if there is a change in code or is this a node issue?

CxRes:
Has this been fixed? 

Or do you want a new issue filed? 

Because the code specifically ignores the `ENOENT` error even for the non-existant symlink.

CxRes:
@paulmillr Can we please discuss this a bit further. I think the fix is not that hard!

As I understand, Chokidar simply waits for a file/directory to be created before issuing the `ready` event. When the item is a symlink, this becomes a two layered process in a way. Chokidar waits for the symlink; on finding a symlink it waits for the directory. 
IMHO, it makes more sense to wait for the symlink item but if the linked directory does not exist it should send an `error` event as it is most likely that the symlink is dangling for good and needs to be fixed separately. With the current behaviour the user has no idea that the wait for `ready` event is because symlink does not exist or whether the symlink is dangling. 

The other option is for user to check for dangling before adding the item, which seems redundant as Chokidar has this information already.

If this seems like a breaking change perhaps this could be introduced under a flag.

paulmillr:
@CxRes nothing to discuss here. We just need the code to get it fixed. And the test cases.

PR #846, opened 2019-05-13
Error and Skip for Dangling Symlinks

CxRes:
This is an obviously incorrect PR. The changes here come with a generous dose of hocus-pocus. I do not fully understand what I am doing but it just works. I am hoping that this will prompt an eventual fix to #469 which now seems to be affecting other people as well.

I have not written any test (only tested the change manually). The test file is so long and dense that it makes me giddy...

Adds a `reportErrorOnDanglingSymlink` option that leads error message when a symlink is found to be dangling. It skips waiting for the underlying path to be created. With this change a dangling symlink will not block the `ready` event, if so desired by the user.


CxRes:
Also, I found the linter complaining in a couple of places (before this PR). I have not touched that for now as it is only cosmetic...

PR #847, opened 2019-05-13
Error and Skip for Dangling Symlinks

CxRes:
This is an obviously incorrect PR (and a second attempt at it). The changes here come with a generous dose of hocus-pocus. I do not fully understand what I am doing but it just works. However I am not sure if it deals with the side effects (e.g. What should happen if the path is recreated or symlink is deleted or linked path is changed? Does the change affect close in some way?). I am hoping that this will prompt an eventual fix to #469 which now seems to be affecting other people as well.

I have not written any test (only tested the change manually). The test file is so long and dense that it makes me giddy. 

The existing tests sometimes pass and sometimes fail. Though I believe the failure is unrelated to changes here:
```
1) chokidar
       fs.watch (non-polling)
         watch glob patterns
           should correctly handle intersecting glob patterns:
     expected spy to have been called exactly twice, but it was called thrice
    spy(add, ~\libs\chokidar\test-fixtures\30\change.txt,
```

Coveralls did complain that the tests are failing in my first attempt, but I do not get those particular errors when testing locally (mind you I am testing on a Windows machine).

Adds a `reportErrorOnDanglingSymlink` option that leads error message when a symlink is found to be dangling. It skips waiting for the underlying path to be created. With this change a dangling symlink will not block the `ready` event, if so desired by the user.

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,6 +167,10 @@ instantiating the watching as chokidar discovers these file paths (before the `r
 * `followSymlinks` (default: `true`). When `false`, only the
 symlinks themselves will be watched for changes instead of following
 the link references and bubbling events through the link's path.
+* `reportErrorOnDanglingSymlinks` (default: `false`) When `followSymlinks: true`, Chokidar silently waits for paths under 
+dangling symlinks to be created. The user might instead want to confine Chokidar to wait for explicitly specified paths only. 
+When set to `true`, the dangling symlink is reported as an error instead and the absence of underlying symlinked path 
+ is ignored.
 * `cwd` (no default). The base directory from which watch `paths` are to be
 derived. Paths emitted with events will be relative to this.
 * `disableGlobbing` (default: `false`). If set to `true` then the strings passed to `.watch()` and `.add()` are treated as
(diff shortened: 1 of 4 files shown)


fictitious:
I have the same problem on my linux laptop, I went ahead and submitted similar, but simple [PR](https://github.com/paulmillr/chokidar/pull/864).

Couple notes about this PR:

1. I don't see how the code added in this PR can cause the Travis error. In fact, with unmodified 3.0.1 code, I've seen the exact same test fail 2 times when I run tests 10 times. 

2. You are adding a new option for skipping dangling symlinks, but there are 2 different implementation for watcher, one in `lib/fsevents-handler` which is used on macOS, another one in `lib/nodefs-handler` which is used everywhere else, and I suspect that the new option you added has no effect when `fsevents-handler` is used. I don't use macOS, but as far as I can see from the code,  on macOS, broken symlinks are handled in the same way as ENOENT errors encountered when accessing other directory entries, that is, silently ignored by default.

paulmillr:
Fixed by subsequent pr.

PR #864, opened 2019-06-25
add call to ready() when an error is handled in _addToNodeFs()

fictitious:
I've encountered #469 on my linux laptop, I can see that when broken symlink is added in [_addToNodeFs](https://github.com/paulmillr/chokidar/blob/master/lib/nodefs-handler.js#L513), `_incrReadyCount()` is called when symlink entry is read in `_handleRead()`, but then ENOENT exception is thrown which bypasses the call to [`ready()`](https://github.com/paulmillr/chokidar/blob/master/lib/nodefs-handler.js#L560) in `_addToNodeFs`.

The result is that `ready` event is never emitted.

This adds calls to `ready()` if the error is handled in `_addToNodeFs()`, in a way similar to [this code](https://github.com/paulmillr/chokidar/blob/master/lib/fsevents-handler.js#L464) in  `fsevens-handler`.

There's also one test for `ready` event with broken symlinks added in this PR.

### DIFF
--- a/lib/nodefs-handler.js
+++ b/lib/nodefs-handler.js
@@ -563,7 +563,10 @@ async _addToNodeFs(path, initialAdd, priorWh, depth, target) {
     return false;
 
   } catch (error) {
-    if (this.fsw._handleError(error)) return path;
+    if (this.fsw._handleError(error)) {
+      ready();
+      return path;
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


paulmillr:
Thank you!

CxRes:
@fictitious Thanks!

@paulmillr Please put out a new release!!!!

PR #877, opened 2019-08-19
refine types: arrays may be readonly

zauberpony:
Hey,

I started using this fine module in a typescript project with lint-rules that enforce the usage of readonly-arrays. Chokidar's types expects mutable arrays, and checking the source-code readonly-arrays should be totally ok, so it's only the types that are too broad.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -17,13 +17,13 @@ export class FSWatcher extends EventEmitter implements fs.FSWatcher {
    * Add files, directories, or glob patterns for tracking. Takes an array of strings or just one
    * string.
    */
-  add(paths: string | string[]): void;
+  add(paths: string | ReadonlyArray<string>): void;
 
   /**
    * Stop watching files, directories, or glob patterns. Takes an array of strings or just one
    * string.
    */
-  unwatch(paths: string | string[]): void;
+  unwatch(paths: string | ReadonlyArray<string>): void;
 
   /**
    * Returns an object representing all the paths on the file system being watched by this
@@ -182,6 +182,6 @@ export interface AwaitWriteFinishOptions {
  * produces an instance of `FSWatcher`.
  */
 export function watch(
-  paths: string | string[],
+  paths: string | ReadonlyArray<string>,
   options?: WatchOptions
 ): FSWatcher;
--- a/types/test.ts
+++ b/types/test.ts
@@ -58,3 +58,9 @@ chokidar
   .on("all", (event: string, path: string) => {
     console.log(event, path);
   });
+
+// watch an (readonly-) array of files
+const files: ReadonlyArray<string> = ["a", "b"];
+const watcher2 = chokidar.watch(files);
+watcher2.add(["c", "d"] as ReadonlyArray<string>);
+watcher2.unwatch(["b"] as ReadonlyArray<string>);


paulmillr:
You've replaced mutable arrays with read-only arrays. But while read-only arrays are subset of generic arrays, generic arrays do not inherit from read-only arrays.

Right now we have a general case. You've reducing this to read-only-arrays only. In fact, on two lines you're explicitly casting types using `as`. This should rarely be done, if ever! I use it only for cases when the compiler is stupid enough to inherit types on its own.

zauberpony:
Just to be sure we're talking about the same thing: "generic arrays" and "mutable arrays" are the same, right?

Since readonly arrays are generic arrays (the simple `[]` in JS) with some methods removed (thus a subset, as you mention) at type-checking-time, I agree that generic arrays do not inherit from readonly-arrays, because they are the exact same at runtime. And even if they would be implemented differently, as long as they behave the same at runtime (duck-typing) they work the same, and TSC is also fine, given it's structural typing.

The current solution is a general case, but could be even more generic, specifing in the type-definition, that the `add`/`watch` methods etc. will just use the subset specified in ReadonlyArray.
In fact, I did not modify your test that use `watcher.add()` with a generic array, since that of course still matches the required structure of a readonly array.

I agree that using `as` in TypeScript to "force cast" something is not good. It's just the question how to best write a literal readonly array, I thought this is test-code with very limited scope, just to ensure that the changes to the type-definitions actually do something.
TypeScript 3.4 introduced `as const` which I would prefer for literal ReadonlyArray, but it seems as if *dtslint* does not (yet?) understand the syntax. Another option is to introduce an explicitly type variable.
But I have no strong feelings about code style here, I you maybe decide to accept readonly arrays nonetheless, I'm happy to change the PR as you request.

paulmillr:
So what i’m trying to understand here is Typescript type hierarchy.

ReadOnlyArray inherits from Array
Array does not inherit from ReadOnlyArray

Which means we will receive type errors when compiling this code: watcher.add([“path”]). Right?

zauberpony:
I see. Well typescript does not use type hierarchy, what is relevant is what I briefly mentioned as "structural typing". TSC sees a type-requirement and checks if the argument fullfills the required interface.
"Inheritance" or "extension" of types in TypeScript is more a tool to reduce required code to write.

Thus
> Which means we will receive type errors when compiling this code: watcher.add([“path”]). Right?

is not right. `watche.add(f: ReadonlyArray<string>)` requires an argument that has some methods, e.g. `find()` — `["path"]` fulfills this requirement, thats enough for TSC to pass the typecheck.

I put together a [simple example](http://www.typescriptlang.org/play/#code/JYOwLgpgTgZghgYwgAgLIE8AicxwMphQCuCYRUKA3gFDJ3IAORYAFANYToBcyIRAtgCNoAGmQA3OABsiEHnyHQAlD3EB7YABMA3LXoBzCGADSndp3kDhUFbyvRdAX2rVQkWIhQYAShDia1ECksHDgAZ0ISMgpkGnpkQxMzDm47RRtLdKcXahgiEFJgQOQwtX4IADF8wsCWTVCeHz8AoJDcCOJScghbBWtYvToKaJBketwAOkTTdBYAFgAmJWyXBECI5H4sMIBJEAi4Ark0NvxIrpiAXgH4plYUzOsxSRkem-j6AHpP5DgoQWAhD+6GQa00EEGyEcIhc8WmyQsaWsvXsUHeH2Qw3IowWAGZdPFnI5dNQ1vswJiIGEiFIKddSuUqgUwEUQCwtphduTDkhlkA) to further demonstrate that.

I also updated the pull-request to be more verbose and show that really generic arrays still work. :)

paulmillr:
Weird. I can’t reopen the PR. Could you create a new one?

zauberpony:
Sure, thanks for taking the time to discuss this issue, I really appreciate it :)

PR #878, opened 2019-08-21
refine types: arrays may be readonly

zauberpony:
chokidar does not modify passed arrays, thus they may also be marked as readonly in type definition.


paulmillr:
Thanks!

paulmillr:
Just a quick clarification from what i've understood.

`ReadonlyArray` is an interface, not a subclass. That's why it's good to use. I've thought it was a subclass. A more modern version of that would be `readonly string[]`, which is only supported by TS 3.4+.

Issue #872, opened 2019-07-30
Can't watch for file creation inside of non-yet-existing directory


pixelastic:
**Describe the bug**
Watching `./foo/bar.txt` when `./foo/` does not yet exists won't trigger any event when `./foo/bar.txt` is finally created.
Having `./foo/` existing when calling `chokidar.watch` or watching `./*/bar.txt` works correctly, though.

**Versions (please complete the following information):**
 - Chokidar version: 3.0.2
 - Node version: v12.6.0
 - OS version: Ubuntu 18.04.2 LTS

**To Reproduce**

```js
const chokidar = require('chokidar');
const fs = require('fs');

// Make sure no ./foo directory currently exists

chokidar.watch('foo/test*', {}).on('all', () => {
  console.log('This will never be called');
});
chokidar.watch('*/test*', {}).on('all', () => {
  console.log('This will be called');
});

fs.mkdirSync('foo');
fs.writeFileSync('foo/test.txt', 'testing 1');
// Both events should fire, but only the second watcher is triggered
```

From reading the tests it seems to be expected behavior (the tests explicitly create directories before watching files), but I can't find any documentation on the subject nor any way to bypass it.

**Expected behavior**
I would expect events to be fired when files are created inside of non-yet-existing directories, no matter how I define my pattern (with a glob, or a static string).

**Additional context**
I'm currently using a hack to force chokidar to treat my static string as a glob by using `(foo|$*)/test*` instead of `foo/test*`. `$*` will never match anything so both `foo` and `(foo|$*)` should match the same things, but it seems to somehow trigger chokidar into the right behavior.

exx8:
Have you tried my pull request?

https://github.com/pixelastic

aleclarson:
This isn't a problem in [filespy](https://github.com/alloc/filespy), an alternative to chokidar.

PR #897, opened 2019-10-09
allow watch for recursive not existing folder

exx8:
Solves #872 

This pull request allows user to watch specific not existing folders.
Was achieved via introducing a new function _mostSpecifiedExistingAncestor
We should consider making  the function public, as users might be interested in this functionality.

### DIFF
--- a/index.js
+++ b/index.js
@@ -770,11 +770,11 @@ _isntIgnored(path, stat) {
  */
 async _getWatchHelpers(path, depth)
 {
-  let watchPath = depth || this.options.disableGlobbing || !isGlob(path) ? path : globParent(path);
-
+  const baseWatchPath = depth || this.options.disableGlobbing || !isGlob(path) ? path : globParent(path);
   const follow = this.options.followSymlinks;
-  watchPath=await this._mostSpecifiedExistingAncestor(watchPath);
-  return new WatchHelper(path, watchPath, follow, this);
+  const existingWatchPath=await this._mostSpecifiedExistingAncestor(baseWatchPath);
+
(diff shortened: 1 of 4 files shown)


paulmillr:
We need tests for that feature. Also, ensure you conform to chokidar code style e.g. `{` locations.

exx8:
I added tests.

index.js
```diff
@@ -765,15 +766,37 @@ _isntIgnored(path, stat) {
...
 
+  /**
+   * Find the most recent (specified ) ancestor path that exists.
+   * @param {Path} path file or directory
+   * @returns {Promise<String>} the most recent ancestor path that exists
+   * @private
+   */
+  async _mostSpecifiedExistingAncestor(path)
+{
+  let father=globParent(path);
```
paulmillr: You need to conform to existing code style. That means using constants, proper indentation etc.

exx8:
I've made the changes that you asked.

exx8:
Ping?

paulmillr:
If i’m reading this correctly, we’ll have a huge performance degradation because of the new IO call for every file. Need to investigate this thoroughly and so some benchmarks.

exx8:
> If i’m reading this correctly, we’ll have a huge performance degradation because of the new IO call for every file. Need to investigate this thoroughly and so some benchmarks.

maybe we should use a flag to enable or disable this functionality?

XhmikosR:
This needs a rebase. Also the curly brackets shouldn't be placed on new lines, there's an extra space in a test message and extra newlines.

PR #900, opened 2019-10-15
requiring a new minimum of fsevents

### LINKED ISSUE #873: Resource starved CPU prevents ready from happening
**Describe the bug**
When starting to watch a directory, if there is another task happening concurrently on the OS, the `ready` event does not fire on MacOS. When the `ready` event doesn't fire, even though `add` events have fired for files, those files do not trigger `change` events when they are changed. I have reproduced this on Sierra as well as in testing in Travis CI.

 - Chokidar version 3.0.2
 - Node version 10.15.3 (tested in 12, 10, and 8 in Travis CI)
 - OS version: MacOS Sierra

**To Reproduce**
I created a gist with some minimal code: https://gist.github.com/catdad/d7f0bff7b47d0796f994174aa078719c

```bash
git clone https://gist.github.com/d7f0bff7b47d0796f994174aa078719c.git
cd d7f0bff7b47d0796f994174aa078719c
npm install

# run watcher alone
node debug.js
# see that files are added and ready is logged

# run watcher through spawn task with other tasks
node spawn.js
# see that files are added but ready is not logged
```

In the `spawn.js` file, if you comment out the second spawn, you can see that even when it is spawned in a child process without other tasks, `ready` is fired as expected.
Steps to reproduce the behavior.

**Expected behavior**
The `ready` event should always fire consistently.

catdad:
Making sure that anyone installing chokidar gets the new version of fsevents which fixes random thread hanging: https://github.com/fsevents/fsevents/pull/290

This helps projects who have a package lock or yarn lock file update their dependencies, as sub-dependencies are not always updated when the existing one matches the semver string.

This resolves #873

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "readdirp": "~3.2.0"
   },
   "optionalDependencies": {
-    "fsevents": "~2.1.0"
+    "fsevents": "~2.1.1"
   },
   "devDependencies": {
     "@types/node": "^12",


paulmillr:
@catdad i’ve forked benchmarks for your watcher — those are pretty good. It’s not a secret that a lean file watcher may eat less RAM or CPU, just keep in mind there are lots of corner cases like different behavior by different editors; for which we have the 2kloc code in place. Thanks!

paulmillr:
We must improve those — for example, by using `recursive: true` option on Windows. That should improve performance a lot.

PR #901, opened 2019-10-16
Update nodejs.yml

XhmikosR:
* move `CI: true` in the workflow root so that it's inherited by all steps
* specify `fail-fast: false`
* split steps
* reindent

Non-whitespace diff: <https://github.com/paulmillr/chokidar/pull/901/files?w=1>

@paulmillr I suppose you skip package-lock.json on purpose?

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -1,27 +1,25 @@
 name: Node CI
-
 on: [push]
+env:
+  CI: true
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


XhmikosR:
Oh, and PRs don't run CI.

XhmikosR:
I'm gonna see tomorrow if I can figure out why tests fail on Windows. That being said, we should have CI for PRs

paulmillr:
Yeah, I don't like lock files, they bring more bugs than needed in some cases. Good for real projects, bad for libs.

PR #906, opened 2019-10-17
Lint codebase

XhmikosR:
* use `const` when possible
* use object destructuring consistently
* use object shorthand
* use the spread operator instead of `Array.from()` and `Object.assign()`
* use template literals
* use arrow functions when possible
* use the strict comparison operators
* remove unneeded else conditions
* remove unneeded escape characters
* remove unused variables

It would be nice if there was stricter linting. JSHint seems to be very much behind compared to ESLint, but this is a decision you should make @paulmillr.

That being said, I tried to not change your style, just tighten things a bit.

If there's something you don't like, let me know and I'll rebase this.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,14 +1,15 @@
 'use strict';
+
 const { EventEmitter } = require('events');
 const fs = require('fs');
 const sysPath = require('path');
+const { promisify } = require('util');
 const readdirp = require('readdirp');
 const anymatch = require('anymatch').default;
 const globParent = require('glob-parent');
 const isGlob = require('is-glob');
 const braces = require('braces');
 const normalizePath = require('normalize-path');
-const { promisify } = require('util');
 
 const NodeFsHandler = require('./lib/nodefs-handler');
 const FsEventsHandler = require('./lib/fsevents-handler');
@@ -88,9 +89,9 @@ const unifyPaths = (paths_) => {
   /**
    * @type {Array<String>}
    */
-  let paths = flatten(arrify(paths_));
+  const paths = flatten(arrify(paths_));
   if (!paths.every(p => typeof p === STRING_TYPE)) {
-    throw new TypeError('Non-string provided as watch path: ' + paths);
+    throw new TypeError(`Non-string provided as watch path: ${paths}`);
   }
   return paths.map(normalizePathToUnix);
 };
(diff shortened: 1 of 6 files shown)


index.js
```diff
@@ -898,7 +898,7 @@ _addPathCloser(path, closer) {
 
 _readdirp(root, opts) {
   if (this.closed) return;
-  const options = Object.assign({type: EV_ALL, alwaysStat: true, lstat: true}, opts);
+  const options = {type: EV_ALL, alwaysStat: true, lstat: true, ...opts};
```
paulmillr: Even though we support 8.16+, some folks go against the advice since chokidar runs on their node 8.x machines. Object spread appeared in 8.10 or so, so we shouldn't use it until we switch to node 10.x+ only

index.js
```diff
@@ -731,7 +731,7 @@ _awaitWriteFinish(path, threshold, event, awfEmit) {
 }
 
 _getGlobIgnored() {
-  return Array.from(this._ignoredPaths.values());
+  return [...this._ignoredPaths.values()];
```
paulmillr: How fast is this when compared to older way?

index.js
```diff
@@ -330,12 +330,12 @@ constructor(_opts) {
     } else if (envLower === 'true' || envLower === '1') {
       opts.usePolling = true;
     } else {
-      opts.usePolling = !!envLower;
+      opts.usePolling = Boolean(envLower);
```
paulmillr: Let's keep the old way, I like it more.

paulmillr:
Feel free to switch this to eslint. I just don't like additional files in repo, so it'll be good to keep config in package.json.

XhmikosR:
We should do this in a separate PR. Any specific config you'd like to use?

index.js
XhmikosR: Oh, OK, I will revert the spread changes then

XhmikosR: Do you think there are more stuff that came in later 8.x versions? I might make a branch with 8.0.0 to test on Ci temporarily to make sure we don't break anything.

paulmillr: https://node.green

XhmikosR: I tested this on Node.js 8.0.0 and it doesn't work already anyway. https://github.com/XhmikosR/chokidar/runs/263880052#step:5:10

So, perhaps we should change the package.json engines version to 8.10?

index.js
XhmikosR: I didn't make any benchmarks. Do you have a script we can use for benchmarking?

XhmikosR:
Alright, I only reverted the `Boolean` change. I kept the spread operator until this is solved

https://github.com/paulmillr/chokidar/pull/906#discussion_r335965880

index.js
paulmillr: Just tested online, no real difference

paulmillr:
That seems to be ready to merge?

XhmikosR:
@paulmillr yup

PR #908, opened 2019-10-17
Tighten JSHint options

XhmikosR:
Lint will pass after #906


XhmikosR:
Yeah I can't figure it out why either. I only managed to get this to work the first time I cloned the repo. Then, all subsequent mocha runs fail, but not always on the same test.

PR #910, opened 2019-10-17
test.js: use `isWindows` and `isMacos` from lib/constants

XhmikosR:


XhmikosR:
@paulmillr BTW judging from readdirp, it seems the Windows image does have git symlinks enabled, which means we can test those on Windows too. It will just fail for normal installations like my Windows VM, because git symlinks are disabled by default on Windows.

PR #915, opened 2019-10-18
Add Coveralls support.

XhmikosR:
Only runs for Node.js 10 and on Ubuntu.

Just testing this, we can use whatever service you like.

The coveralls action has one downside: you will get a comment in PRs even if you have them disabled.

Anyway, let me know if you want this. If you do, please add the repo in the service of your choice and I will adapt this patch.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -22,3 +22,9 @@ jobs:
       - run: npm install
       - name: Run tests
         run: npm test
+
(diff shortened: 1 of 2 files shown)


paulmillr:
gh actions is fine

XhmikosR:
Yeah, but GH Actions is the CI not a coverage service.

XhmikosR:
Unless you don't want such a service, then this patch is moot.

PR #916, opened 2019-10-18
Switch to ESLint

XhmikosR:
@paulmillr this is just a basic config, that extends from` eslint:recommended`. I tried to match the current style and patterns.

Feel free to push here to fix the remaining lint errors. I still believe the test failures are a race async condition.

Also, personally I wouldn't put the eslint config in package.json, but you said you prefer this.

### DIFF
--- a/example.js
+++ b/example.js
@@ -1,3 +1,5 @@
+'use strict';
+
 global.watcher = require('.').watch('.', {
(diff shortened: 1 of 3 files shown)


XhmikosR:
@paulmillr can you have a look at the errors? Feel free to push to this branch. :)

PR #924, opened 2019-10-21
Update rimraf to v3.0.0

XhmikosR:
https://github.com/isaacs/rimraf/compare/v2.7.1...v3.0.0

Changes:

- Add `--preserve-root` option to executable (default true)
- Drop support for Node.js below version 6


paulmillr:
We should either use readdirp for recursive removal, or fs.rmdir with recursive: true. I don’t think the module is needed anymore.

XhmikosR:
`fs.rmDir` recursive is experimental and only available on Node.js >= 12.x.

paulmillr:
11c76f7

PR #933, opened 2019-10-26
Update devDependencies.

XhmikosR:
* dtslint to 0.9.9
* eslint 6.6.0
* mocha 6.2.2


XhmikosR:
BTW it seems dtslint does not run on CI. That's OK, after we have caching on actions, we can change this and cache the download folder.

Issue #967, opened 2020-01-04
Symlinks are always watched ignoring matching pattern and `followSymlinks` option


ClassicOldSong:
**Describe the bug**

Symlinks are always watched ignoring matching pattern and `followSymlinks` option.

**Versions (please complete the following information):**
 - Chokidar version 3.3.1
 - Node version v12.12.0
 - OS version: MacOS 10.15

**To Reproduce:**

Create a file named `test.aaa`
Create a file named `test.bbb`
Create a symlink to `test.bbb` named `test.ccc`

```javascript
const chokidar = require('chokidar')

const watcher = chokidar.watch('**/*.aaa')

const log = console.log.bind(console)

watcher
  .on('add', path => log(`File ${path} has been added`))
  .on('change', path => log(`File ${path} has been changed`))
  .on('unlink', path => log(`File ${path} has been removed`))
```

**Expected behavior**
Only `test.aaa` should be watched, but `test.ccc` is watched as well.

**Additional context**
Project: https://github.com/TheNeuronProject/ef.qt
Related code: https://github.com/TheNeuronProject/ef.qt/blob/master/watcher.js#L113
When project includes symlinks, they are no matter how watched and caused issue on code compiling .

paulmillr:
The backlog is going to start from zero as a preparation for v4 release. v4 would bring massive rewrite to the table and drop most dependencies. All issues are being closed as preparation for v4 release.

In the future, only issues with enough community support would be considered.

See issue 1195 for more info. Thank you.

PR #969, opened 2020-01-11
Partial fix for broken symlink handling

jaens:
Solves my immediate issue, but there are other bugs remaining related to broken symlinks.

I was not able to proceed any further due to lack of understanding because unfortunately the code does not document its data structures (eg. `_symlinkPaths`) or what invariants in general it expects to hold.


jaens:
Not sure why the "should properly match glob patterns that include a symlinked dir" test occasionally fails.

paulmillr:
yeah, I don't think this would handle it properly without introducing new issues. see #967, we can't even get current behavior right.

PR #992, opened 2020-03-13
Make close always return promises

### LINKED ISSUE #991: Calling close on a closed watcher returns the watcher, not a Promise
**Describe the bug**

Calling `close` on a closed watcher does not return a Promise as documented, but instead returns the watcher itself.

 - Chokidar version: every version after https://github.com/paulmillr/chokidar/commit/124ed20af3c74ea577d0477d138da345ff451883
 - Node version: v12.16.1
 - OS version: Ubuntu 19.04

**To Reproduce:**
```js
const fw = chokidar.watch('.')

console.log(fw.close() instanceof Promise); // true
console.log(fw.close() === fw); // true
```

**Expected behavior**
Every call to `close` should return a Promise. All of these Promises should resolve at the same time. Probably easiest to achieve by saving the Promise returned the first time in a private member and returning it here: https://github.com/paulmillr/chokidar/blob/3af3b3df84ddfff11f51fd8756a09479bf1e69cd/index.js#L486

raphinesse:
Fixes #991

### DIFF
--- a/index.js
+++ b/index.js
@@ -483,7 +483,7 @@ unwatch(paths_) {
  * @returns {Promise<void>}.
 */
 close() {
-  if (this.closed) return this;
+  if (this.closed) return this._closePromise;
   this.closed = true;
 
   // Memory management.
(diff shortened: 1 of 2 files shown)


raphinesse:
The CI failures seem unrelated to this change.

PR #1003, opened 2020-04-26
add `safe-edit` handling for Linux, fixes #591

### LINKED ISSUE #591: rename raw event breaks further events
.watch seems to break when there's a 'rename' event on Linux and no more events are fired from there on.
Vim fires this on wq and Git on checkout, for example.

The current workaround I'm using seems to fix this:
```javascript
this.watcher.on('raw', (event, path, {watchedPath}) => {
            if (event === 'rename') {
                this.watcher.unwatch(watchedPath);
                this.watcher.add(watchedPath);
            }
        });
```
It seems this might be easily patched out by verifying that the renamed file has the same name and re-attach the listener automatically.

pkit:
will re-add the watcher on inode change in `fs.watch` mode on Linux

bonus: fixed strange, pretty infrequent race in `fs.watchFile` test case `should emit matching dir events`

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -59,3 +59,4 @@ exports.IDENTITY_FN = val => val;
 
 exports.isWindows = platform === 'win32';
 exports.isMacos = platform === 'darwin';
+exports.isLinux = platform === 'linux';
--- a/lib/nodefs-handler.js
+++ b/lib/nodefs-handler.js
@@ -6,6 +6,7 @@ const { promisify } = require('util');
 const isBinaryPath = require('is-binary-path');
 const {
   isWindows,
+  isLinux,
   EMPTY_FN,
   EMPTY_STR,
   KEY_LISTENERS,
@@ -356,8 +357,7 @@ _handleFile(file, stats, initialAdd) {
   // if the file is already being watched, do nothing
   if (parent.has(basename)) return;
 
-  // kick off the watcher
-  const closer = this._watchWithNodeFs(file, async (path, newStats) => {
+  const listener = async (path, newStats) => {
     if (!this.fsw._throttle(THROTTLE_MODE_WATCH, file, 5)) return;
     if (!newStats || newStats.mtimeMs === 0) {
       try {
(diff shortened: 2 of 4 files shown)


pkit:
Oh, I see choke full of flaky tests.

lib/constants.js
```diff
@@ -59,3 +61,4 @@ exports.IDENTITY_FN = val => val;
 
 exports.isWindows = platform === 'win32';
 exports.isMacos = platform === 'darwin';
+exports.isLinux = platform === 'linux';
```
paulmillr, line 64: What about BSD?

lib/nodefs-handler.js
```diff
@@ -69,6 +72,15 @@ const delFromSet = (main, prop, item) => {
 
 const isEmptySet = (val) => val instanceof Set ? val.size === 0 : !val;
 
+// will return inode number on linux (if file exists) or undefined
+function inode (path) {
+  if (!isLinux) return;
+  try {
+    const stat = fs.statSync(path)
```
paulmillr, line 79: We can't use synchronous IO, switch to async method here.

lib/nodefs-handler.js
```diff
@@ -102,7 +114,14 @@ const FsWatchInstances = new Map();
  * @returns {fs.FSWatcher} new fsevents instance
  */
 function createFsWatchInstance(path, options, listener, errHandler, emitRaw) {
+  const originalInode = inode(path)
+
   const handleEvent = (rawEvent, evPath) => {
+    // linux-specific: if inode returns not-null we need to emit inode change event
+    const newInode = inode(path)
```
paulmillr, line 121: I don't think we can do this. Chokidar is extremely performance-conscious. We count every `stat` call. Consider a 600,000-file dir tree. You're adding two `stat`s, which can degrade performance massively.

paulmillr:
Thanks for the pull request. Unfortunately it's not as simple as it looks. See my `stat` comment.

lib/nodefs-handler.js
pkit: It should work for individual file watchers only. Directories are not in the path.
I don't see any benchmark suites in the repo though...

lib/constants.js
pkit: The problem exists for `inotify` only (AFAIK) which is strictly Linux

lib/nodefs-handler.js
paulmillr: well, 600k-file tree. I'm using readdirp benchmark https://github.com/paulmillr/readdirp/blob/master/examples/bench.js - we should create something similar here.

pkit: well, each event does a stat when listener is called [here](https://github.com/paulmillr/chokidar/blob/b67d208fbd7314dd77135adbb84e2b9c509f4345/lib/nodefs-handler.js#L364)
and also there is a stat at the startup time
I think I can reuse these. Which means no new stats will be added (apart from obvious re-stat of an existing path to fix the bug itself)

this readdir bench is just pumping stuff into a directory, my PR will not change that path at all.
it's enabled only when people watch each individual file, which is a pretty bad idea for 600k dir, it will be slow in any case.

anyway, I think I can probably prove that the perf difference is negligible
but it will need some time to get samples of real workloads

paulmillr: Nice! Will wait for your results then.

pkit: @paulmillr okay, now I don't use any new stats. 
it plugs in a completely different place

pkit:
@paulmillr bottom line, current PR works fine for me, doesn't add any overhead apart from obvious re-install of a listener for a new inode.
I have no idea how to fix the flaky tests though, the system looks too fragile.

paulmillr:
Спасибо!

Issue #955, opened 2019-11-29
Improper handling of broken symlinks, causes UnhandledPromiseRejectionWarning


jaens:
When watching a directory with `followSymlinks` off, if the directory contains links for which the target does not exist, chokidar blows up with:

```
UnhandledPromiseRejectionWarning: Error: ENOENT: no such file or directory, stat <path of symlink>
```

This is due to calling `fsrealpath` on the link, which fails (since the link target does not exist) here:
https://github.com/paulmillr/chokidar/blob/7c79e45f83b52084d03940ed882e0279e1c23b5f/lib/nodefs-handler.js#L416

Note the node.js documentation for `fs.realpath`:
> Asynchronously computes the canonical pathname by resolving ., .. and **symbolic links**.

Perhaps using the `path` module would be more appropriate there.

jaens:
Looks like `fsrealpath` is also being used in a similar manner here:
https://github.com/paulmillr/chokidar/blob/7c79e45f83b52084d03940ed882e0279e1c23b5f/lib/nodefs-handler.js#L607

ruruiye:
same problem here

paulmillr:
pull requests are welcome

nicks:
fwiw, this breaks create-react-app's webpack server hard, because it treats these warnings as errors and crashes the server

The error looks like this:

```
/home/nick/src/tilt/web/node_modules/react-scripts/scripts/start.js:19
  throw err;
  ^

[Error: ENOENT: no such file or directory, stat '/home/nick/src/tilt/web/src/.#Statusbar.tsx'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'stat',
  path: '/home/nick/src/tilt/web/src/.#Statusbar.tsx'
}
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

Here's the stack trace to the call that throws the error.

```
    at internal/util.js:297:30
    at new Promise (<anonymous>)
    at internal/util.js:296:12
    at NodeFsHandler._handleSymlink (/home/nick/src/tilt/web/node_modules/watchpack/node_modules/chokidar/lib/nodefs-handler.js:433:30)
    at ReaddirpStream.<anonymous> (/home/nick/src/tilt/web/node_modules/watchpack/node_modules/chokidar/lib/nodefs-handler.js:486:23)
    at ReaddirpStream.emit (events.js:321:20)
[...]

PR #1010, opened 2020-06-02
handle promise rejection when a symlink's target does not exist. Fixe…

nicks:
…s https://github.com/paulmillr/chokidar/issues/955


jaens:
Is there a reason why this was not merged?

At least it fixes #955 while also not adding **more** test failures (some of the tests fail on current main branch anyway on my machine).

paulmillr:
reason is simple: it's hard to understand which use cases this will break. We have millions of users, and the tests aren't really comprehensive, since they fail on CIs. 

I know that this is not a reason to halt progress, but we need more reviews of each PR etc.

PR #1025, opened 2020-08-03
handle network path `//server` correctly

### LINKED ISSUE #895: Watching network drive '\\server\folder' does not work
Hi
I have a problem where chokidar does not report changes to network shared folder, i.e. `\\\\file.server.net\\dev\\7500`. Using `fs.writeFile()` on that folder works fine and a file gets created by Node. I'm under Win10.
My config is below. I tried both with `usePolling` true and false, no change.
I don't get any errors, just changes to network folder don't get reported. Watching a folder on local drive C:\db works fine. Watching that network drive using drive letter path format actually works too `s:\\dev\\7500`. I'd prefer to use server path notation in case drive mapping letters get changed in future, or some users in the company have different letters mapped.

Do you see anything amiss in my setup?

```
fs.writeFile('\\\\file.server.net\\dev\\7500\\test.txt', 'this file gets created OK', (err) => {
  if (err) throw err
})

const chokidarOptions = {
  persistent: true,
  // ignored: "*.txt",
  ignoreInitial: false,
  followSymlinks: true,
  cwd: ".", 
  disableGlobbing: false,
  usePolling: true, // tried with false as well
  interval: 1000,
  binaryInterval: 3000,
  alwaysStat: false,
  depth: 99,
  awaitWriteFinish: {
    stabilityThreshold: 2000,
    pollInterval: 100,
  },
  ignorePermissionErrors: true, // watch files that don't have read permissions
  atomic: true, // or a custom 'atomicity delay', in milliseconds (default 100)
};

const optionsFileMonitor = {
  watchFolders: ["\\\\file.server.net\\dev\\7500", "s:\\dev\\7500", "C:/db"]
};

const fileMonitor = chokidar
  .watch(optionsFileMonitor.watchFolders, chokidarOptions)
  .on("all", (event, epath) => {
    console.log(event, path.resolve(epath));
})
```

whyboris:
In current `master` branch, `chokidar` fails to recognize paths starting with `//` or `\\` for example:

`//StoragePC/DrivePool/movies` is a valid network path, but `toUnix` method would replace `//` with `/` making the path invalid.

More details: https://github.com/paulmillr/chokidar/issues/895#issuecomment-660518084

Closes #895 

Tested with my app. PR successfully resolves the problem: https://github.com/whyboris/Video-Hub-App/issues/456#issuecomment-664038303

Unit test results unchanged from `master`:
```
  192 passing (1m)
  12 pending
  1 failing     <-- should emit `unlinkDir` and `add` when dir is replaced by file:
```

please let me know if you prefer the constant name to be `DOUBLE_SLASH` instead of `SLASH_SLASH` 👍

### DIFF
--- a/index.js
+++ b/index.js
@@ -34,6 +34,7 @@ const {
   REPLACER_RE,
 
   SLASH,
+  SLASH_SLASH,
   BRACE_START,
   BANG,
   ONE_DOT,
@@ -96,10 +97,19 @@ const unifyPaths = (paths_) => {
   return paths.map(normalizePathToUnix);
 };
 
+// If SLASH_SLASH occurs at the beginning of path, it is not replaced
+//     because "//StoragePC/DrivePool/Movies" is a valid network path
 const toUnix = (string) => {
   let str = string.replace(BACK_SLASH_RE, SLASH);
+  let prepend = false;
+  if (str.startsWith(SLASH_SLASH)) {
+    prepend = true;
+  }
   while (str.match(DOUBLE_SLASH_RE)) {
     str = str.replace(DOUBLE_SLASH_RE, SLASH);
+  }
+  if (prepend) {
+    str = SLASH + str;
   }
   return str;
 };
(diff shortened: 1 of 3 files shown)


nicograef:
Thanks a lot to both of you @whyboris @paulmillr ! 🙏

jmeinke:
It would be great if you could release this fix asap - thank a lot!!

Issue #1040, opened 2020-10-09
[macos] watchPath seems to be false


RoXuS:
**Describe the bug**

I serve lot of files (about 430) and certain files are not watched for some obscur reason.
I added some log before `fsevent.watch`, for example for these two files (they are symlink):

- /Users/julienrousseau/Documents/insight/monorepo/node_modules/@doubletrade/dt-queries-utils/dt-queries-table.js
- /Users/julienrousseau/Documents/insight/monorepo/node_modules/@doubletrade/dt-app/dt-notification.ts

The watch folder is `/Users/julienrousseau/Documents/insight/monorepo/node_modules/@doubletrade` after:
````
  if (couldConsolidate(parentPath)) {
    watchPath = parentPath;
  }
````
But only change on `dt-notification.ts` emit change event.

I don't understand why...

If I comment this:
````
  if (couldConsolidate(parentPath)) {
    watchPath = parentPath;
  }
`````

All my files are correctly watched.

**Versions (please complete the following information):**
 - Chokidar version ^3.4.0
 - Node version 12/13/14
 - OS version: Last macos

**To Reproduce:**

I don't know ...

RoXuS:
More precision.

Two same folder but one is symlink:
`/Users/julienrousseau/Documents/insight/monorepo-new/packages` -> Launch event on `/Users/julienrousseau/Documents/insight/monorepo-new/packages/xx/x.js`
`/Users/julienrousseau/Documents/insight/monorepo-new/node_modules/@doubletrade` -> Doesn't launch event on `/Users/julienrousseau/Documents/insight/monorepo-new/node_modules/@doubletrade/xx/x.js`

RoXuS:
Can ben solved by using the real path instead of symlink path, is it a good choice?
https://github.com/paulmillr/chokidar/pull/1041/files

paulmillr:
the new tests at my mac keep failing. @RoXuS can you check this?

RoXuS:
hi @paulmillr  on my mac it is ok?

Do you do something that special??
![Uploading image.png…]()

paulmillr:
@RoXuS can you test it without .only?

RoXuS:
yeah I get this error
![image](https://user-images.githubusercontent.com/3296671/111866718-bff7df00-896f-11eb-8a8c-b5fcedb54b99.png)

Tests doesn't seem to be unitary?
I check how to do that.

I have some other errors, you too?

[...]

PR #1041, opened 2020-10-10
Use realPath instead of path on watchers

RoXuS:
Attempt to fix https://github.com/paulmillr/chokidar/issues/1040

### DIFF
--- a/lib/fsevents-handler.js
+++ b/lib/fsevents-handler.js
@@ -104,7 +104,8 @@ const createFSEventsInstance = (path, callback) => {
  * @returns {Function} closer
  */
 function setFSEventsListener(path, realPath, listener, rawEmitter) {
-  let watchPath = sysPath.extname(path) ? sysPath.dirname(path) : path;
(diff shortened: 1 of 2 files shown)


LarsDenBakker:
@paulmillr is this a good solution for this issue?

paulmillr:
The problem with chokidar is that since it's used so much, even a change like this needs to be tested manually in apps etc. So just need to test it...

RoXuS:
hey! Any news on this PR? All of my mate have this issues on mac...

Can I help on something?

paulmillr:
would be great if you had a test for the case

RoXuS:
@paulmillr, I added the test, it was hard to reproduce from scratch.
I have to create more than 10 folders to reach the `consolidateThreshold` then I add symlink on each and I listen these symlinks with chokidar.
When the `consolidateThreshold` is reached, the watch folders are set to the parent then the `change` events aren't listen anymore.

If you comment my change on `fsevents-handler.js` you will see that the event is never fired and the test failed.


If can I do something, tell me!

RoXuS:
thx @paulmillr, me and my team will be happy ^^

PR #1063, opened 2021-01-12
Bugfix/#1061

CuddlySheep:
**Problem:**

The basic bug was the same as already fixed in Issue #1042.
However, the bugfix solved the problem only for relative paths, but not for symlink paths (part of the path is a symlink).

The reason for this was the **incorrect** assumption that:
`fs.realpath(path) != path ==> path is a symlink path`

**Correct assumption**:
`fs.realpath(path) != path ==> path is a symlink path or a relative`

**Solution:**
`fs.realpath(path) != fs.resolve(path) ==> path is a symlink path`

Also, it was a bug to remember the resolved path instead of link-path, because with this is used internally.


CuddlySheep:
After merging this bugfix it would be very nice to provide a new version of chokidar 3.5.1.

Thank you for the good cooperation.

PR #1067, opened 2021-01-23
Update package.json

AlAyoub:
glob-parent was flagged as of this morning with high severity security risk. The recommendation is to downgrade to 3.0.0, 2.0.0, or 1.0.0.


efussi:
The fix https://github.com/gulpjs/glob-parent/pull/36 has been merged, and the dependency should be updated to `5.1.1`.

PR #1070, opened 2021-01-28
Fix missing events on IBM i

### LINKED ISSUE #1069: Events not emitted on IBM i platform
**Describe the bug**

Events not emitted at all on IBM i platform.

 - Chokidar version 3.2.1
 - Node version 10.23.1
 - OS version: IBM i 7.4

**To Reproduce:**
To reproduce, attempt to use the chokidar library on the IBM i platform. Events seem to be not emitted at all. 
This causes several things to not work properly, for instance the `--watch` parameter of PM2. 

**Expected behavior**
Events are expected to be emitted as documented. 

**Additional context**
I believe the underlying root cause is simply that the underlying `fs.watch()` support [is not available on the IBM i platform](https://nodejs.org/docs/latest/api/fs.html#fs_availability). A simple workaround is to use polling mode via `CHOKIDAR_USEPOLLING=1`, but I'll work on a small PR that defaults the behavior to polling mode for this platform.

ThePrez:
Resolves #1069 

Pretty small change, but initial testing shows positive results with chokidar on the IBM i platform. I will perform more verification and un-draft in a day or two, but opening as a draft PR in the meantime for comments. Thanks in advance for consideration!

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ On MacOS, chokidar by default uses a native extension exposing the Darwin
 implementations like `kqueue` available on most \*nix platforms. Chokidar still
 does have to do some work to normalize the events received that way as well.
 
-On other platforms, the `fs.watch`-based implementation is the default, which
+On most other platforms, the `fs.watch`-based implementation is the default, which
 avoids polling and keeps CPU usage down. Be advised that chokidar will initiate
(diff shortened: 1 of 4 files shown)


paulmillr:
Thanks

PR #1077, opened 2021-03-05
Limit node version to <= 12

scinos:
This PR adds a limitation on node versions. It should ensure `chokidar@2.x` doesn't get accidentally installed in unsupported Node versions.

[Reference](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#engines)

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,9 @@
     "lib/",
     "types/index.d.ts"
   ],
+  "engines": {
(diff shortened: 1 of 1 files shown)


paulmillr:
I'm afraid there are tons of folks who still install it on newer version and we can't do anything here. It's like hundreds of thousands of them

Issue #1024, opened 2020-07-31
Only build FSEvents when using MacOS


nicholaschiang:
**Describe the bug**

NPM and Yarn log a warning message saying that the FSEvents build was skipped when running on Linux and Windows machines (this is a duplicate of #460). This is expected.

Ideally, though, `chokidar` would be smart enough to avoid this warning (when it is expected) by *only* installing and building FSEvents when the user is on MacOS.

**Versions (please complete the following information):**
 - Chokidar version: 2.1.8
 - Node version: 14.7.0
 - OS version: Ubuntu 18.04.2

**To Reproduce:**

1. Get Linux or Windows machine.
2. Install `chokidar` using Yarn v2 (or NPM, the warning should still be there).
3. See warning message:
```
$ yarn install
➤ YN0000: ┌ Resolution step
➤ YN0000: └ Completed in 0.52s
➤ YN0000: ┌ Fetch step
➤ YN0000: └ Completed in 0.99s
➤ YN0000: ┌ Link step
➤ YN0062: │ fsevents@patch:fsevents@npm%3A2.1.3#builtin<compat/fsevents>::version=2.1.3&hash=495457 The platform linux is incompatible with this module, building skipped.
➤ YN0062: │ fsevents@patch:fsevents@npm%3A1.2.13#builtin<compat/fsevents>::version=1.2.13&hash=495457 The platform linux is incompatible with this module, building skipped.
➤ YN0000: └ Completed in 1.97s
➤ YN0000: Done with warnings in 3.71s
```

**Expected behavior**

FSEvents should only be installed, built, and used when the user is on MacOS (thus avoiding this warning to begin with).

paulmillr:
I don't see a way to solve this in chokidar. Report this to NPM or Yarn. We specify supported platforms in fsevents package.json, these bundlers should understand this.

cwadrupldijjit:
FYI for any who arrive here looking at the same sort of issue, this issue does have a bug report on NPM, though I don't know that anybody from the NPM team itself has looked at it, since I don't see any comments written by one of them yet.  I don't much like the workarounds for the issue, either, but I figure that I could stick a link to the issue here for any who are interested:
https://github.com/npm/cli/issues/2921

PR #1083, opened 2021-03-20
fix some tests by adding delay and moving some tests

RoXuS:
hi @paulmillr,

By adding some delays and moving some tests, it is a little better.

My tests are ok but some continue to failed.
The failed are not link to my fix. Some failed seems to be random...

The #1024 continue to failed on my mac with or without my changes, but with my changes it is a little better.

The "close" test involve lot of errors but if it move to the end of the file, it is ok...

Can you test this fix on your mac?


RoXuS:
thx @paulmillr, if I can help on something let me know.

PR #1171, opened 2021-11-16
Replace braces dependency with brace-expansion

tpacent:
### Why?

1. This project uses braces for the sole purpose of glob pattern expansion. Braces is overly complex for the use case.
1. Braces is obsolete -- last updated 3 years ago
1. The braces package depends on the ridiculous is-number package

My PR suggests replacing braces with brace-expansion.  
It is similar in popularity (156 github stars vs 150 for braces, 26M downloads vs 48M for braces), there is some commit activity so it is maintained, it has exactly 1 dependency.

### DIFF
--- a/index.js
+++ b/index.js
@@ -8,7 +8,7 @@ const readdirp = require('readdirp');
 const anymatch = require('anymatch').default;
 const globParent = require('glob-parent');
 const isGlob = require('is-glob');
-const braces = require('braces');
+const expand = require('brace-expansion');
 const normalizePath = require('normalize-path');
 
 const NodeFsHandler = require('./lib/nodefs-handler');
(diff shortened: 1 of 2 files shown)


paulmillr:
Can you check whether our sub-dependencies use braces? Because it doesn't make sense to only replace ours

tpacent:
> Can you check whether our sub-dependencies use braces? Because it doesn't make sense to only replace ours

Mocha is a devDependency using chokidar which in turn uses braces. I guess this will resolve itself in time.
```
chokidar@3.5.2
└─┬ mocha@7.2.0
  └─┬ chokidar@3.3.0
    └── braces@3.0.2
```

PR #1195, opened 2022-01-18
Version 4: Typescript, ESM, glob removal, drop node <18

paulmillr:
- [ ] Release readdirp v4
- [ ] Changelog
- [ ] Release chokidar v4


package.json
```diff
@@ -1,46 +1,41 @@
...
-    "anymatch": "~3.1.2",
-    "braces": "~3.0.2",
-    "glob-parent": "~5.1.2",
-    "is-binary-path": "~2.1.0",
-    "is-glob": "~4.0.1",
-    "normalize-path": "~3.0.0",
-    "readdirp": "~3.6.0"
+    "is-binary-path": "2.1.0",
+    "normalize-path": "3.0.0",
+    "readdirp": "github:paulmillr/readdirp"
```
lydell, line 19: This reduction in dependencies is awesome! 😍 💯 

Are the versions pinned on purpose?

paulmillr: yes, to ensure packages don't get hacked and replaced by malware

sw360cab:
@paulmillr is this alpha release blocked ?
Is there any plan to release it? Maybe some contribution needed.

Thanks a lot for your great work.

paulmillr:
Nah -- someone just needs to find some time to debug it.

paulmillr:
@43081j there are a few issues:

- we are using mocha which depends on old chokidar. Maybe we should replace it with something else? For the record, I have made micro-should, but it does not have beforeEach / afterEach. perhaps worth manually adding "beforeEach" code into functions and seeing if it works
- there are some warnings after npm install that need to be fixed
- eslint seems to be outdated

paulmillr:
@benmccann how flexible are you in terms of ESM-onliness?

43081j:
> * we are using mocha which depends on old chokidar. Maybe we should replace it with something else? For the record, I have made micro-should, but it does not have beforeEach / afterEach. perhaps worth manually adding "beforeEach" code into functions and seeing if it works

we should be able to use node's test runner now (since 18.x at least, maybe even 16.x)

> * there are some warnings after npm install that need to be fixed
> * eslint seems to be outdated

these are probably both solved by us updating and cleaning up some dependencies. ill take a look

paulmillr:
I have made many hybrid packages manually and it's nothing hard. It's just launching tsc two times with different configs. See https://github.com/paulmillr/jsbt

`tshy` seems like too much deps. They even use chokidar -- wtf? why?

43081j:
as a dev dependency. the important thing tshy does is create the package `exports` map for you

i don't mind either way as long as things like [arethetypeswrong](https://arethetypeswrong.github.io/) say our package is setup correctly (if it doesn't, its likely ts consumers will get build errors in some configurations)

paulmillr:
> create the package exports map for you

1. Exports map can be written by hand in a few minutes, or using a small shell script that lists directory files: https://github.com/paulmillr/noble-hashes/blob/b930aa959dfb95a936096b5ac79a3dcfbdab2332/package.json#L46-L70
2. I have been writing that by hand, however, since we're going full-esm, I don't think it's necessary. We only have one export. Even for hybrid, it's very easy to write.
3. Overall exports are mess and not all envs support them properly. Ideally they'll just be removed in place of absolute paths e.g. noble-hashes/sha2.js instead of noble-hashes/sha2

43081j:
typescript, node and many bundlers will now assert that the imported path is in the export map so we should set it up

but im fine with us doing it by hand too, whatever we can all agree on 👍  i wouldn't export extensionless paths etc, i'd pretty much export the on-disk files (aliasing for cjs if we dual-package, and exporting them directly)

paulmillr:
We also need to make this ESM and upgrade https://github.com/paulmillr/readdirp/

43081j:
> We also need to make this ESM and upgrade https://github.com/paulmillr/readdirp/

i have a branch where i've tried to remove `readdirp` since node itself should be capable by now (with `recursive: true`)

though it has a few test failures so i was planning on trying to fix those

what do you think?

unrelated - the same failure as my last few PRs happened again here. seems to be windows specific 🤔 i'd like to solve that one somehow

paulmillr:
ideally we'll just use node built-ins, ofc.

Try readdirp-ing 400K-file directory and seeing how's the speed with both solutions. I imagine nodejs is still shit. Keep in mind that readdirp uses streams etc.

43081j:
thats true, i'll try it out on a larger project if i can find one some place 👍

43081j:
i did some benchmarking, and it seems like readdirp recursively visits some symlinked directories, whereas node won't

in a test repo:

> readdirp - Found 122112 items
> node - Found 32967 items

it seems the missing items are already part of the `32967` (symlinked by pnpm in this case)

in a directory that doesn't hit that problem, node seems faster:

```
┌─────────┬────────────┬─────────┬────────────────────┬──────────┬─────────┐
│ (index) │ Task Name  │ ops/sec │ Average Time (ns)  │ Margin   │ Samples │
├─────────┼────────────┼─────────┼────────────────────┼──────────┼─────────┤
│ 0       │ 'readdirp' │ '9'     │ 105280229.69999999 │ '±9.32%' │ 10      │
│ 1       │ 'node'     │ '41'    │ 23874190.62790698  │ '±3.59%' │ 43      │
└─────────┴────────────┴─────────┴────────────────────┴──────────┴─────────┘
```

paulmillr:
Symlinks are definitely important.

43081j:
> Symlinks are definitely important.

im not saying node is ignoring them. readdirp is duplicating them from what i can tell

> readdirp - Found 122112 items
> node - Found 32967 items

the extra 89,145 entries seem to be files in `node_modules` we already visited (already accounted for in the 32,967).

symlinks certainly are traversed via node

benmccann:
> @benmccann how flexible are you in terms of ESM-onliness?

Sorry I missed this question, but don't worry about me. I'm fine with whatever you decide either way

43081j:
@paulmillr if you didn't see, it looks like the windows failures fail in master too

it does seem like it could be a bug but we could at least release a pre-release for now. that would be nice i think (`4.0.0-pre.0` or something, in a `next` tag on npm)

the functionality seems pretty sound and i've been through the diff with master, which shows our only changes really are:

- removal of globs
- ignore logic

so should be good, and any problems we see are likely already problems (that we should fix)

paulmillr:
I would still prefer to fix tests for v4. They have been broken for many years.

Do you think we can do it?

43081j:
i agree

its a busy couple of days coming up but ill take a look again when i can and see if i can figure it out

43081j:
@paulmillr it is a bug, i've figured it out roughly but not how to fix it

when we add a node to the watcher, we `stat` the node here:
https://github.com/paulmillr/chokidar/blob/c8c3ff2dc2b47d206225b93ced71e4004aabefdb/src/nodefs-handler.ts#L622

when we write a file (`add.txt`), we cause a `rename` fs event

this results in us calling the `handleDir` watcher here:
https://github.com/paulmillr/chokidar/blob/c8c3ff2dc2b47d206225b93ced71e4004aabefdb/src/nodefs-handler.ts#L587-L592

that starts readdirp going on the directory:
https://github.com/paulmillr/chokidar/blob/c8c3ff2dc2b47d206225b93ced71e4004aabefdb/src/nodefs-handler.ts#L483-L488

which finds `add.txt` and adds the node:
https://github.com/paulmillr/chokidar/blob/c8c3ff2dc2b47d206225b93ced71e4004aabefdb/src/nodefs-handler.ts#L518

we then reach this `stat` call (**which is async**):
https://github.com/paulmillr/chokidar/blob/c8c3ff2dc2b47d206225b93ced71e4004aabefdb/src/nodefs-handler.ts#L622

**meanwhile, a change event fires** - nothing really comes of this because we haven't finished our `stat` of the file, so haven't added a listener yet

**when the test passes** - a second change event fires at this point, and there's still no file listener so we do not fire a `CHANGE` event

**now the `stat` finishes** and adds the file listener to the set of listeners

**when the test fails** - the second change event fires at this point, and there _is_ a file listener so we fire a `CHANGE` event

43081j:
🤔 

is it possible this is just wrong:
https://github.com/paulmillr/chokidar/blob/c8c3ff2dc2b47d206225b93ced71e4004aabefdb/src/nodefs-handler.ts#L385

specifically `at <= mt`:

- if the access time is before the modified time, it means we haven't accessed it since it was modified, so this is a change
- if the access time is _the same as_ the modified time, surely it implies it hasn't changed?

if i change this to `at < mt`, the tests pass

millette:
Related: #1336

benjamingwynn:
Will this release address #1225, or will another issue have to be created?

paulmillr:
New release drops support for globs. So, it would address it by setting it "out of scope". As in: all glob patterns would become unsupported. You will need to write regexps, or filtering functions.

PR #1197, opened 2022-01-22
handle MustScanSubDirs

### LINKED ISSUE #1196: Chokidar does not handle kFSEventStreamEventFlagMustScanSubDirs properly, calls unlink callback
**Describe the bug**

If the system is flooded with too many fsevents (for example, you are watching a directory that undergoes a huge number of changes all at once, and the user code to process these events is slow), a buffer of events can fill up and trigger a kFSEventStreamEventFlagUserDropped event:

> The kFSEventStreamEventFlagUserDropped or kFSEventStreamEventFlagKernelDropped flags may be set in addition to the kFSEventStreamEventFlagMustScanSubDirs flag to indicate that a problem occurred in buffering the events (the particular flag set indicates where the problem occurred) and that the client must do a full scan of any directories (and their subdirectories, recursively) being monitored by this stream. If you asked to monitor multiple paths with this stream then you will be notified about all of them. Your code need only check for the kFSEventStreamEventFlagMustScanSubDirs flag; these flags (if present) only provide information to help you diagnose the problem.

When this happens...

fsevents calls the watch callback on the root directory with flags=3 (kFSEventStreamEventFlagMustScanSubDirs & kFSEventStreamEventFlagUserDropped), and then calls the callback for every file and subdirectory.

When chokidar processes the event on the root directory, it calls fsevents.getInfo and compares the type to the previous type. If the new type does not match the old type, it calls the callback with the unlink event. 

Chokidar code:
https://github.com/paulmillr/chokidar/blob/master/lib/fsevents-handler.js#L151
https://github.com/paulmillr/chokidar/blob/master/lib/fsevents-handler.js#L331-L335

The problem is that fsevents.getInfo().type will be undefined if the flags=3. (Relevant: https://github.com/fsevents/fsevents/issues/372) See the implementation here: https://github.com/fsevents/fsevents/blob/master/fsevents.js#L40-L54

Because "directory" !== undefined, chokidar incorrectly calls the unlink event on the root directory, and user applications cannot distinguish between a true unlink and the kFSEventStreamEventFlagMustScanSubDirs event.

 - Chokidar version: 3.5.2
 - Node version: 12.20.1
 - OS version: MacOS 12.1

**To Reproduce:**

Create a chokidar watch on a directory. The callback should do something expensive like busy-loop. Flood the directory with frequent updates to a large number of files.

**Expected behavior**
Chokidar sends a special event signaling that this behavior has happened, or at very least does not call unlink on the root directory.

**Additional context**
We are working on a large mono repo and watch the root directory to transpile our code. We have a separate tool that typechecks the entire repo and outputs .d.ts files to a .gitignored directory within the repo. When running these tools at the same time, typescript generates a large number of watches that are somewhat expensive to process (need to re-transpile the code) which repros this issue.

-Submitted by Airtable

MarcCelani-at:
If the system is flooded with too many fsevents (for example, you are watching a directory that undergoes a huge number of changes all at once, and the user code to process these events is slow), a buffer of events can fill up and trigger a kFSEventStreamEventFlagMustScanSubDirs event.

See the [following documentation](https://developer.apple.com/documentation/coreservices/1455361-fseventstreameventflags/kfseventstreameventflagmustscansubdirs):

> The kFSEventStreamEventFlagUserDropped or kFSEventStreamEventFlagKernelDropped flags may be set in addition to the kFSEventStreamEventFlagMustScanSubDirs flag to indicate that a problem occurred in buffering the events (the particular flag set indicates where the problem occurred) and that the client must do a full scan of any directories (and their subdirectories, recursively) being monitored by this stream. If you asked to monitor multiple paths with this stream then you will be notified about all of them. Your code need only check for the kFSEventStreamEventFlagMustScanSubDirs flag; these flags (if present) only provide information to help you diagnose the problem.

When this happens, fsevents.info() returns an undefined `type`, which chokidar incorrectly interprets as an unlink. This PR explicitly checks for kFSEventStreamEventFlagMustScanSubDirs before calling fsevents.info and ignores events of this type.

Relevant: fsevents/fsevents#372

Fixes #1196

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -24,6 +24,7 @@ exports.FSEVENT_DELETED = 'deleted';
 exports.FSEVENT_MOVED = 'moved';
 exports.FSEVENT_CLONED = 'cloned';
 exports.FSEVENT_UNKNOWN = 'unknown';
+exports.FSEVENT_FLAG_MUST_SCAN_SUBDIRS = 1;
 exports.FSEVENT_TYPE_FILE = 'file';
 exports.FSEVENT_TYPE_DIRECTORY = 'directory';
 exports.FSEVENT_TYPE_SYMLINK = 'symlink';
--- a/lib/fsevents-handler.js
+++ b/lib/fsevents-handler.js
@@ -37,6 +37,7 @@ const {
   FSEVENT_MOVED,
   // FSEVENT_CLONED,
   FSEVENT_UNKNOWN,
+  FSEVENT_FLAG_MUST_SCAN_SUBDIRS,
   FSEVENT_TYPE_FILE,
   FSEVENT_TYPE_DIRECTORY,
   FSEVENT_TYPE_SYMLINK,
@@ -48,9 +49,6 @@ const {
   EMPTY_FN,
   IDENTITY_FN
 } = require('./constants');
-
-// See kFSEventStreamEventFlagMustScanSubDirs in FSEvents documentation
-const MUST_SCAN_SUBDIRS_FLAG = 1;
 
 const Depth = (value) => isNaN(value) ? {} : {depth: value};
 
@@ -151,7 +149,7 @@ function setFSEventsListener(path, realPath, listener, rawEmitter) {
       rawEmitter,
       watcher: createFSEventsInstance(watchPath, (fullPath, flags) => {
         if (!cont.listeners.size) return;
-        if (flags & MUST_SCAN_SUBDIRS_FLAG) return;
+        if (flags & FSEVENT_FLAG_MUST_SCAN_SUBDIRS) return;
         const info = fsevents.getInfo(fullPath, flags);
         cont.listeners.forEach(list => {
           list(fullPath, flags, info);
--- a/lib/fsevents-handler.js
+++ b/lib/fsevents-handler.js
@@ -49,6 +49,9 @@ const {
   IDENTITY_FN
 } = require('./constants');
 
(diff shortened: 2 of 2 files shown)


paulmillr:
Thanks for the PR. Do you think it should be better handled instead of `fsevents` package?

MarcCelani-at:
I'm assuming you meant "inside of the `fsevents` package"

How would you propose fsevents handle this? I see fsevents as a way to expose the fsevents api to node programs. I would want it to be as true as possible to the actual underlying api. So it should definitely pass the exact flags it receives back to chokidar (in other words, I would not want fsevents to drop the callback).

I suppose that you could argue that getInfo(flags).type should return 'directory' here. But then, shouldn't MacOS set the kFSEventStreamEventFlagItemIsDir flag when it sends this event (it doesn't)? I'd argue that the fsevents module is just mirroring whatever MacOS does, and that's a pretty good thing for a module that is exposing something not otherwise available in nodejs.

I would say that fsevents should at least make more clear in the .d.ts file that getInfo().type can be undefined. I filed fsevents/fsevents#372 for that.

MarcCelani-at:
@paulmillr , let me know how best to proceed here

MarcCelani-at:
@paulmillr, any thoughts here?

paulmillr:
fsevents maintainer opinion needs to be included, we'll wait for it

beorn:
One year and no answer from fsevents - except someone suggested to use libuv, which apparently uses similar base tech, and which all modern node versions have.

PR #1198, opened 2022-01-27
Remove unused devDependencies

XhmikosR:


package.json
```diff
@@ -30,11 +31,10 @@
     "dtslint": "^3.3.0",
     "eslint": "^7.0.0",
     "mocha": "^7.0.0",
-    "nyc": "^15.0.0",
     "rimraf": "^3.0.0",
     "sinon": "^9.0.1",
     "sinon-chai": "^3.3.0",
-    "typescript": "~4.4.3",
+    "typescript": "^4.4.3",
```
glensc, line 37: typescript doesn't follow semver, so `"~4.4.3",` would be accurate regarding breaking changes.

glensc:
The PR does a bit more than just "Remove unused devDependencies"

43081j:
Back in 2022.

Much of the project has changed since then, including how we build and which dependencies we pull in. 

How did you find yourself here? Was there a problem?

PR #1200, opened 2022-01-28
fix argument name in ReadMe

jubilee2:


jubilee2:
the first argument is path, second one is event @paulmillr

Issue #923, opened 2019-10-21
Fix CI


XhmikosR:
We really need a stable CI :/

1. fix/ignore the lint errors (the require-atomic-updates error might be a FP, needs a closer look from someone more familiar with the codebase)
2. fix tests

I was able to get tests to pass consistently on my Windows VM if I skip this test:

https://github.com/paulmillr/chokidar/blob/00329fce7182b8a1a9dba2c2851f797c9a15d3a7/test.js#L198

```
  181 passing (60s)
  14 pending
```

On CI, it just keeps failing consistently, though https://github.com/XhmikosR/chokidar/runs/268382878

If anyone has any ideas please submit a patch :)

PR #1228, opened 2022-07-08
Fix for a race condition on Linux

### LINKED ISSUE #1112: Race condition when watching dirs leads to missed files
**Describe the bug**

There is a basic race condition when setting up a watch on a directory, in `nodefs-handler.js`:  First the directory contents are scanned, and then a `fs.watch` watcher is set up.  Any file created between those two operations won't be seen, until there's another operation that causes the `fs.watch` to trigger and the directory is re-scanned.

 - 3.5.1
 - 10.21.0
 - Linux

**To Reproduce:**

First add this delay, which doesn't cause the race, just makes it a lot easier to trigger.

```diff
--- a/lib/nodefs-handler.js
+++ b/lib/nodefs-handler.js
@@ -556,6 +556,9 @@ async _handleDir(dir, stats, initialAdd, depth, target, wh, realpath) {
       if (this.fsw.closed) return;
     }
 
+    console.error("race!");
+    await new Promise(res => setTimeout(res, 5000));
+
     closer = this._watchWithNodeFs(dir, (dirPath, stats) => {
       // if current directory is removed, do nothing
       if (stats && stats.mtimeMs === 0) return;
```

Then use something like this

```js
const fs = require('fs').promises;

function sleep(n) {
    return new Promise(res => setTimeout(res, n * 1000));
}

(async () => {
    try {
        const dir = await fs.mkdtemp('/tmp/foo-');

        const w = chokidar.watch(dir);
        w.on('all', (event, path) => {
            console.log("-- EVENT:", event, path);
        });
        w.on('ready', () => {
            console.log("-- EVENT: ready");
        });

        await sleep(2);
        console.log(`creating ${dir}/1`);
        await fs.writeFile(`${dir}/1`, "hi");

        await sleep(2);
        console.log(`creating ${dir}/2`);
        await fs.writeFile(`${dir}/2`, "hi");

        await sleep(2);
        console.log("nothing?");

        await sleep(10);
        console.log(`creating ${dir}/3`);
        await fs.writeFile(`${dir}/3`, "hi");

        await sleep(2);
        process.exit(0);
    } catch (e) {
        console.error(e);
    }
})();
```

**Actual output**

Events for adding files 1 and 2 don't happen until file 3 is created much later:

```
-- EVENT: addDir /tmp/foo-A2kUIi
race!
creating /tmp/foo-A2kUIi/1
creating /tmp/foo-A2kUIi/2
-- EVENT: ready
nothing?
creating /tmp/foo-A2kUIi/3
-- EVENT: add /tmp/foo-A2kUIi/1
-- EVENT: add /tmp/foo-A2kUIi/2
-- EVENT: add /tmp/foo-A2kUIi/3
```

**Expected output**

Events for adding files 1 and 2 should happen right before or after
`ready`:

```
-- EVENT: addDir /tmp/foo-A2kUIi
race!
creating /tmp/foo-A2kUIi/1
creating /tmp/foo-oh0HPl/2
-- EVENT: ready
-- EVENT: add /tmp/foo-A2kUIi/1
-- EVENT: add /tmp/foo-A2kUIi/2
nothing?
creating /tmp/foo-A2kUIi/3
-- EVENT: add /tmp/foo-A2kUIi/3
```

dividedmind:
Sometimes, when another process just happened to create a file
when we were setting up a watch, the new file wouldn't be noticed.

To prevent the event not being generated in this case, schedule
a directory rescan immediately after setting up a watch.

Fixes #1112

---
Also, another commit here to fix brittle tests. It might be related to #923

### DIFF
--- a/index.js
+++ b/index.js
@@ -136,6 +136,8 @@ const getAbsolutePath = (path, cwd) => {
 
 const undef = (opts, key) => opts[key] === undefined;
 
+const NODE_MAJOR_VERSION = Number(process.versions.node.split('.')[0]);
+
 /**
  * Directory entry.
  * @property {Path} path
@@ -950,6 +952,17 @@ _readdirp(root, opts) {
       stream = undefined;
     }
   });
+
+  if (NODE_MAJOR_VERSION < 10) {
+    // On Node 8 Readable.push() throws when the stream is destroyed
+    // even though the documentation says it should be silently ignored.
+    const superPush = stream.push.bind(stream);
+    stream.push = function (...args) {
+      if (this.destroyed) return false;
+      return superPush(...args);
+    }
+  }
+
   return stream;
 }
 
--- a/lib/nodefs-handler.js
+++ b/lib/nodefs-handler.js
@@ -570,6 +570,11 @@ async _handleDir(dir, stats, initialAdd, depth, target, wh, realpath) {
 
       this._handleRead(dirPath, false, wh, target, dir, depth, throttler);
     });
+
+    // schedule a rescan of the directory in case something
+    // appeared while we were setting up the watch
+    if (!(this.fsw.usePolling || target || this.fsw.closed))
+      this._handleRead(dir, false, wh, target, dir, depth, throttler);
   }
   return closer;
 }
--- a/test.js
+++ b/test.js
@@ -487,6 +487,29 @@ const runTests = (baseopts) => {
       expect(spy.args[0][1]).to.be.ok; // stats
       rawSpy.should.have.been.called;
     });
+    if (!(baseopts.usePolling || baseopts.useFsEvents)) context('when a race happens', () => {
+      beforeEach(() => {
+        const stub = sinon.stub(fs, 'watch');
+        stub.callsFake((filename, ...rest) => {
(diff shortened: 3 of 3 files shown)


dividedmind:
Looks like the fix is breaking some tests on non-linux platforms. Do you think it's a good idea to conditionally do this on Linux only? Or do you think the failures are unrelated? I don't have access to either platform so I can't say for sure, but the test suite is quite brittle. The one fix I included here seemed to be enough to get it to pass robustly on my system, but there might be more dragons there.

paulmillr:
The nodefs-handler is not used on macos, so i'm not sure how to fix it. Maybe in the same way in fsevents-handler.

dividedmind:
That's a good point, I forgot. In this case I think it's fair to say that these test failures must be unrelated.

dividedmind:
I changed it to only do the readdir if polling is enabled and added a workaround for a Node 8 bug. I wanted to take a look at more Node 8 test problems, but then I realized it's over two years past EOL and it's probably not worth the trouble. I verified tests pass robustly on Node >= 14 on Linux. The MacOS and Windows test failures seem unrelated, so I think this is good to merge.

PR #1253, opened 2022-11-21
Fix no events for symlinks with invalid targets

nathanael-ruf:
Even with `followSymlinks: false` no events (e.g. 'add') were emitted for links with a target that doesn't exist.
Not sure about the changes / why we even need the realpath if `followSymlinks: false`. Also tests seem to fail (at least locally).
This is more like a draft PR so far (maybe someone can take over or suggest changes).

Steps to repro:
```
const chokidar = require('.');
chokidar.watch('tmp', {
    usePolling: true,
    followSymlinks: false
}).on('all', (e, p) => console.log(e, p));
```

```
ln -s does-not-exist link
```

### DIFF
--- a/test.js
+++ b/test.js
@@ -2133,7 +2133,7 @@ const runTests = (baseopts) => {
         await delay(300);
         await write(testSubDirFile, '');
         await delay(300);
-
+        
         chai.assert.deepStrictEqual(events, [
           `[ALL] addDir: ${sysPath.join('test-fixtures', id, 'test')}`,
           `[ALL] addDir: ${sysPath.join('test-fixtures', id, 'test', 'dir')}`,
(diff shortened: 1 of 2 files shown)


test.js
```diff
@@ -2122,7 +2133,7 @@ const runTests = (baseopts) => {
         await delay(300);
         await write(testSubDirFile, '');
         await delay(300);
-        
+
```
nathanael-ruf, line 2136:
```suggestion
        
```

test.js
```diff
@@ -2163,7 +2174,7 @@ const runTests = (baseopts) => {
         await fs_mkdir(testSubDir);
         await write(testSubDirFile, '');
         await delay(300);
-        
+
```
nathanael-ruf, line 2177:
```suggestion
        
```

Issue #1191, opened 2022-01-10
There is no CVE-2021-35065, Chokidar is not vulnerable


paulmillr:
CVE-2021-35065 only applies to glob-parent 5.1.1 and 6.0.0, it **does not apply** to 5.1.2 which we are using. **glob-parent 5.1.2 is not vulnerable**. We will not update to 6.0 because chokidar 3 needs to support nodejs v8.

If your tool tells you chokidar is vulnerable, report issues to your build tool. White Source Software is particular piece of shit since it does not do proper checks.

https://github.com/github/advisory-database/pull/531 https://github.com/github/advisory-database/pull/533

dzzk:
> We will not update to 6.0 because chokidar 3 needs to support nodejs v8.
Do you mean NodeJS Carbon v8x.x the last time updated on 2019-12-17 or you are talking about v8 engine itself?

And thank you for the quick response you did.

paulmillr:
yes, we will still support node from 2019, because chokidar is used by tens of millions of users and some of them cannot upgrade easily their child deps.

dzzk:
Isn't it a nice idea if I will make pr for chokidar 4 with all possible changes for greater nodejs versions within your repo? I do not want to waste GitHub environment for nothing and increase entropy in Universe, but need to have 6.0.2 version somewhere.

What do you think?

paulmillr:
Chokidar 4 will need a typescript rewrite probably. I don't think it matters without big changes

paulmillr:
#1195

mtarnawa:
@paulmillr 
https://nvd.nist.gov/vuln/detail/CVE-2021-35065#range-8736225

Just to clarify - the NVD record was modified couple of days ago to include also 5.1.2 hence the inflow of new issues. I understand that the official stance is still that 5.1.2 is not vulnerable?

mtarnawa:
> @mtarnawa read the first post one more time. It doesn't matter who posts this stuff.

I did read it. As well as the relevant issue in glob-parent project from the year back. I was surprised to see CPEs were updated suddently in the last couple of days that's why I wanted to reconfirm - it's no longer about vendor of security tool xyz being overzealous, but it's now being flagged in NVD which is bad, but from what you are saying - still incorrect.

llpaul:
[...]

PR #1259, opened 2023-01-09
chore: upgrade gblob-parent to fix CVE-2021-35065

edgar-berlinck:


edgar-berlinck:
https://github.com/paulmillr/chokidar/issues/1191

PR #1261, opened 2023-01-10
Bump glob-parent to 6.0.2

PraveenAnaparthi:
there is a vulnerability in the current version so updating it to the latest version.


paulmillr:
😂

PR #1288, opened 2023-09-06
fix readyCount logic

JLHwung:
This PR fixes `readyCount` logic.

Previously the `this._readyCount` will be doubled whenever `watcher.add()` is called, I believe it should have been only doubling the ready count change (hence adding the `paths.length` again) rather than the total counts. After all, calling `watcher.add()` for three times should yield the same `readyCount` as calling it one time with all 3 different paths.

I added a test for this fix. The test hangs forever on macOS because `readyCount` is too large. With the new fix, the `ready` event is emitted as expected.


JLHwung:
The current CI errors seem to be unrelated, since they are breaking on the main branch as well.

PR #1307, opened 2024-02-14
Fix filter pattern not being applied to files inside symlink subdirectories of the watched path

samuelneff:
This PR resolves issue #967 and applies the pattern filter to files nested inside symlinks. 

No change in expected behavior or usage. New passing test added.

### DIFF
--- a/index.js
+++ b/index.js
@@ -202,16 +202,21 @@ class DirEntry {
 const STAT_METHOD_F = 'stat';
 const STAT_METHOD_L = 'lstat';
 class WatchHelper {
-  constructor(path, watchPath, follow, fsw) {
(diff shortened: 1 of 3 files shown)


samuelneff:
I'll address those issues, assuming they're real issues (lint at least probably is). 👍

samuelneff:
@paulmillr It looks like none of the errors are actually related to my PR but are existing issues. This particular one is with DTS Lint and a missing function.

paulmillr:
As i’ve mentioned, we can’t really merge new code unless we fix the old tests.

Otherwise some stuff could get broken and we wouldn’t even know.

Thanks for the pr.

samuelneff:
Thanks for the update. I missed that specific message about the tests. I'll work on those next. 👍

PR #1318, opened 2024-05-25
test: switch to node test runner

43081j:
Switches to using the built-in node test runner and drops mocha.

Temporarily also introduces a `TEST_TIMEOUT` as `--test-timeout` is only available in 20.x and above, but we run CI in 18.x.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,10 @@
   "version": "4.0.0-alpha",
   "homepage": "https://github.com/paulmillr/chokidar",
   "author": "Paul Miller (https://paulmillr.com)",
-  "contributors": ["Paul Miller (https://paulmillr.com)", "Elan Shanker"],
+  "contributors": [
+    "Paul Miller (https://paulmillr.com)",
+    "Elan Shanker"
+  ],
   "engines": {
     "node": ">= 18"
   },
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -4,7 +4,10 @@
   "version": "4.0.0-alpha",
   "homepage": "https://github.com/paulmillr/chokidar",
   "author": "Paul Miller (https://paulmillr.com)",
-  "contributors": ["Paul Miller (https://paulmillr.com)", "Elan Shanker"],
+  "contributors": [
```
43081j, line 7: npm automatically reformats this FYI

43081j:
looks like tests are failing on windows quite often 😞 

its late, ill have a look some other time soon

paulmillr:
all good?

43081j:
there's some flakiness/timing problem with the tests on windows when we use node's runner

i have a pretty hectic weekend but ill try pick it up soon and let you know

43081j:
seems we have a passing test run!

on my fork, CI still fails every now and then for this test:
https://github.com/43081j/chokidar/actions/runs/9337536632/job/25699552795#step:5:84

its a failed assertion rather than a timeout, so i will try investigate it more as the `changeSpy` definitely shouldn't have been called

PR #1322, opened 2024-06-11
chor:bump braces version to 3.0.3

airene:
see:https://github.com/advisories/GHSA-grv7-fg5c-xmjg


paulmillr:
https://stackoverflow.com/questions/22343224/whats-the-difference-between-tilde-and-caret-in-package-json

PR #1335, opened 2024-06-30
fix: consider atime == mtime to be a non-change

43081j:
Changes the file listener to treat `atime === mtime` as a non-change.

Currently, if these are equal, we fire a `CHANGE` event which causes bugs on windows (as windows emits two change events when you first write).

### DIFF
--- a/src/nodefs-handler.ts
+++ b/src/nodefs-handler.ts
@@ -372,17 +372,15 @@ export default class NodeFsHandler {
     if (parent.has(basename)) return;
 
     const listener = async (path, newStats) => {
-      console.log({path, newStats});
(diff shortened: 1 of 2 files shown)


43081j:
nevermind, looks like it breaks ubuntu now and fixes windows... 🤦‍♂️ 

a problem for another day

43081j:
saw you merged the v4 branch @paulmillr , awesome 🎉 

ill try pick this one up tomorrow. if we can fix this bug we're pretty much ready to go i think

i could do with help on it though, especially given ill be out travelling from weds so my time will drop for a bit

PR #1339, opened 2024-07-30
chore: remove leftover console.logs

43081j:
leftover from some debugging in the v4 branch

### DIFF
--- a/src/nodefs-handler.ts
+++ b/src/nodefs-handler.ts
@@ -372,12 +372,10 @@ export default class NodeFsHandler {
     if (parent.has(basename)) return;
 
     const listener = async (path, newStats) => {
-      console.log({path, newStats});
       if (!this.fsw._throttle(THROTTLE_MODE_WATCH, file, 5)) return;
       if (!newStats || newStats.mtimeMs === 0) {
         try {
           const newStats = await stat(file);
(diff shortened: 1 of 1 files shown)


paulmillr:
FYI the v4 branch is WIP locally here and ongoing a huge refactoring. It's much more complicated then initially thought, but getting there.

43081j:
is there somewhere easy to reach you so we could catch up on this stuff?

would be happy to help, and it'd be nice to know what you're changing etc. i don't want to end up duplicating work but want to help get it over the line still

paulmillr:
https://github.com/paulmillr/chokidar/tree/alpha 

Check out this document: https://gist.github.com/paulmillr/43e375be993b9a86bc43888e837345b3

43081j:
that's a great write up 🙏 

if you need any help along the way, just let me know

super happy you found the time to get back onto this

paulmillr:
@43081j check out the code for `v4.ts` - any help in refactoring this is needed.

43081j:
if you can make a draft PR, I can leave comments too. I think that'd be helpful so I could point out any areas I'm looking at, assuming I do some

paulmillr:
https://github.com/paulmillr/chokidar/pull/1342

PR #1342, opened 2024-08-01
Massive rewrite, simplification

paulmillr:


src/v4.ts
```diff
@@ -0,0 +1,1201 @@
...
+    // NOTE: this is only used correctly in readdir for now.
+    // How it should work:
+    // - we process some event
+    // - same event happens in parallel multiple times
+    // - when we processed first event, we look at last throttled event
+    // - start processing it
+    // How it works now (except readdir):
+    // - we process first event (first change)
+    // - there is multiple parallel changes which we throw away
+    // - all changes which happened when we processed first event is lost
```
43081j, line 522: fixing this would probably also solve the windows bug

just a recap of what that is/was:

writing a file in windows often fires two events, while it fires one on linux/mac

43081j:
how are things going with this @paulmillr?

will be out travelling soon but if you need a hand with anything, shout up and ill try chip in

paulmillr:
I will upload new commit soon and then pause for a bit. It's too complicated. Feel free to code it.

43081j:
ill take a look once you push 👍 

if you can leave any comments where you think stuff needs figuring out too, that'd be helpful

would be so nice for us to get this over the line some day

paulmillr:
@43081j I think we should release current work as v4 and then the further refactor in this pull request as v5 when it's ready.

I want to make chokidar massively compatible for v4. Could you help me to understand what's the lowest version we support and why? Perhaps we'll use commonjs + esm hybrid for v4.

43081j:
that makes sense

we really should fix the windows bug at least, though. since it is an actual bug

to remind you what that is:

- write a file
- we start a readdirp stream and start adding children
- meanwhile, windows fires a second event
- the readdirp didn't finish yet, so this second event thinks the file hasn't yet been seen and fires extra events

something along the lines of that. if we can fix it, i think v4 is all good

as for lowest node version - it should still work fine on anything `>=18`. if we want to support lower, we could try. i guess the limitation will be what node features we use, though maybe we're not relying on any newer ones yet

paulmillr:
It seems like readdirp is used with depth: 0 which means it's not recursive. So we don't even need it directly. Maybe we should just remove it.

43081j:
I noticed this too and have a branch where I switched to using node's recursive flag instead (if at all). But it had a few failures since it switches the logic from streams to promises so is a chunky rewrite 

I agree we probably should do it at some point though

paulmillr:
readdirp provides a bunch of features:

- fileFilter, dirFilter
- lstat
- maybe something else

just keep this in mind. I think we can do it quickly.

paulmillr:
>we really should fix the windows bug at least, though. since it is an actual bug

Did it also happen in v3?

43081j:
I'm pretty sure it did since we haven't changed the logic around there 

We changed tests that started to pick it up but it existed before afaict

paulmillr:
@43081j can you do a sanity check on main branch? I want to release it. Perhaps i've missed something?

Also, what were the improvements that you've added for the rewrite?

paulmillr:
The CI seems to be failing because of timeouts, need to fix that as well

PR #1349, opened 2024-09-13
fix: allow second argument to be `undefined`

benmccann:


benmccann:
Thanks! Any chance I can get a 4.0.1? We'd love to upgrade, but would need this first

paulmillr:
I want to wait a day or so to fix other discovered bugs.

you can use {} as workaround today.

Issue #1355, opened 2024-09-16
v4: types and jsdoc do not align


benmccann:
**Describe the bug**

The `index.d.ts` file has:
```
    /**
     * Close watchers and remove all listeners from watched paths.
     * @returns {Promise<void>}.
     */
    close(): Promise<void> | undefined;
```

**Versions (please complete the following information):**
 - Chokidar version: 4.0.0

**To Reproduce:**

Open `index.d.ts`

**Expected behavior**

Types and docs should match

**Additional context**

This prevents `wrangler` from upgrading because it expects a `Promise<void>` to be returned:
```
src/dev/use-esbuild.ts(226,4): error TS2322: Type '() => Promise<void> | undefined' is not assignable to type '() => Promise<void>'.
```

43081j:
I think it should be `Promise<void>`

It may be worth annotating the return type in the source so we're certain it's always a promise (i.e to catch where we forgot to return one)

PR #1356, opened 2024-09-18
fix: remove jsdoc types and improve internal types

talentlessguy:
should fix https://github.com/paulmillr/chokidar/issues/1355

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -52,7 +52,7 @@ export type FSWInstanceOptions = BasicOpts & {
 };
 
 export type ThrottleType = 'readdir' | 'watch' | 'add' | 'remove' | 'change';
-export type EmitArgs = [EventName, Path, any?, any?, any?];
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -525,9 +523,8 @@ export class FSWatcher extends EventEmitter {
 
   /**
    * Close watchers and remove all listeners from watched paths.
-   * @returns {Promise<void>}.
    */
-  close() {
+  close(): Promise<void> | undefined {
```
talentlessguy, line 527: @benmccann `this._closePromise` may be undefined, should we return undefined as well, or use a non-null assertion?

benmccann: I didn't get to look at this before it was merged, but to answer your question I think `Promise<void>` is the correct return type here and we should not return `undefined`. I've opened a new issue to discuss it: https://github.com/paulmillr/chokidar/issues/1357

PR #1359, opened 2024-09-19
fix: always return promise in close()

43081j:
Updates the `close` method to always return a `Promise` rather than having a mixed return type.

We do not need to check `closed` since we only care if the close logic/promise has already started, so switching the condition should be enough.

cc @benmccann


benmccann:
Thank you!

PR #1360, opened 2024-09-19
fix: remove fsevents from docs and JSDoc

43081j:
Removes `fsevents` from the current documentation and the JSDoc/types.

### DIFF
--- a/src/handler.ts
+++ b/src/handler.ts
@@ -140,7 +140,7 @@ const FsWatchInstances = new Map<string, FsWatchContainer>();
  * @param listener main event handler
  * @param errHandler emits info about errors
  * @param emitRaw emits raw event data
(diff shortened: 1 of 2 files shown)


src/handler.ts
```diff
@@ -140,15 +140,15 @@ const FsWatchInstances = new Map<string, FsWatchContainer>();
  * @param listener main event handler
  * @param errHandler emits info about errors
  * @param emitRaw emits raw event data
- * @returns new fsevents instance
+ * @returns {NativeFsWatcher}
```
paulmillr, line 143: can we remove all `@returns`?

43081j: we could

I'll see if I can remove them all in a separate PR 👍

PR #1363, opened 2024-09-20
improve internal types a little bit

talentlessguy:
as a bonus remove `flatten` util and use a native method instead

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -39,6 +39,12 @@ type BasicOpts = {
   // ioLimit?: number; // Limit parallel IO operations (CPU usage + OS limits)
 };
 
+export type Throttler = {
+  timeoutObject: any;
+  clear: () => any;
+  count: number;
+}
+
 export type ChokidarOptions = Partial<
   BasicOpts & {
     ignored: Matcher | Matcher[];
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -142,20 +148,8 @@ function anymatch(matchers: Matcher[], testString: string | undefined): boolean
   return matchPatterns(patterns, testString);
 }
 
-const flatten = (list: any[], result = []) => {
```
talentlessguy, line 145: `Array#flat` is supported since Node v12

https://node.green/#ES2019-features-Array-prototype--flat--flatMap-

bluwy: In the places where you call `.flat()`, I think you need to pass `.flat(Infinity)` to replicate the old behaviour. Though with that said, the types doesn't allow nested arrays in the first place so I'm not sure if flattening is needed, but maybe better to keep for compat?

talentlessguy: well `any[]` can as well be `any[][]` so it doesn't differ

and `flatten` is an internal util that was only used once so I don't think setting infinite depth makes sense here

bluwy: I don't mean `any[]`, but where it was being used and called here:

image

Arguably flattening with any level doesn't make sense given the types, but if we ignore the types and assume people could nest with any depth, then `.flat(Infinity)` matches the previous behaviour.

PR #1381, opened 2024-10-29
feat: strongly type event emitter methods

### LINKED ISSUE #1372: Missing types for event callbacks
- Chokidar  v4.0.1
- TS v5.5

```ts
import chokidar from "chokidar";

chokidar.watch("src").on('change', (path) => {
  //                                 ^ this is any for me in v4
});
```

v3 types have `implements fs.FSWatcher` on `FSWatcher`  which made things better typed it seems

43081j:
Adds strong types to the event emitter interface such that `on` and similar methods can infer parameters.

For example:

```ts
// these are strongly typed and will give hints in editors etc
watcher.on('ready', () => {});
watcher.on('error', (err) => {});

// this will fail
watcher.on('nonsense', () => {});
```

Fixes #1372

cc @paulmillr

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -13,6 +13,7 @@ import {
   EMPTY_FN,
   STR_CLOSE,
   STR_END,
+  WatchHandlers,
 } from './handler.js';
 
 type AWF = {
@@ -58,7 +59,8 @@ export type FSWInstanceOptions = BasicOpts & {
 };
 
 export type ThrottleType = 'readdir' | 'watch' | 'add' | 'remove' | 'change';
-export type EmitArgs = [EventName, Path | Error, any?, any?, any?];
+export type EmitArgs = [Path | Error, Stats?];
+export type EmitArgsWithName = [EventName, ...EmitArgs];
 export type MatchFunction = (val: string, stats?: Stats) => boolean;
 export interface MatcherObject {
   path: string;
@@ -295,6 +297,17 @@ export class WatchHelper {
   }
 }
 
+export interface FSWatcherKnownEventMap {
+  [EV.READY]: [];
+  [EV.RAW]: Parameters<WatchHandlers['rawEmitter']>;
+  [EV.ERROR]: Parameters<WatchHandlers['errHandler']>;
+  [EV.ALL]: [EventName, ...EmitArgs];
+}
+
+export type FSWatcherEventMap = FSWatcherKnownEventMap & {
+  [k in Exclude<EventName, keyof FSWatcherKnownEventMap>]: EmitArgs;
+};
+
 /**
  * Watches files & directories for changes. Emitted events:
  * `add`, `addDir`, `change`, `unlink`, `unlinkDir`, `all`, `error`
@@ -303,7 +316,7 @@ export class WatchHelper {
  *       .add(directories)
  *       .on('add', path => log('File', path, 'was added'))
  */
-export class FSWatcher extends EventEmitter {
+export class FSWatcher extends EventEmitter<FSWatcherEventMap> {
   closed: boolean;
   options: FSWInstanceOptions;
 
@@ -315,13 +328,13 @@ export class FSWatcher extends EventEmitter {
   _watched: Map<string, DirEntry>;
 
   _pendingWrites: Map<string, any>;
-  _pendingUnlinks: Map<string, EmitArgs>;
+  _pendingUnlinks: Map<string, EmitArgsWithName>;
   _readyCount: number;
   _emitReady: () => void;
   _closePromise?: Promise<void>;
   _userIgnored?: MatchFunction;
   _readyEmitted: boolean;
-  _emitRaw: () => void;
+  _emitRaw: WatchHandlers['rawEmitter'];
   _boundRemove: (dir: string, item: string) => void;
 
   _nodeFsHandler: NodeFsHandler;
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -58,7 +59,8 @@ export type FSWInstanceOptions = BasicOpts & {
 };
 
 export type ThrottleType = 'readdir' | 'watch' | 'add' | 'remove' | 'change';
-export type EmitArgs = [EventName, Path | Error, any?, any?, any?];
+export type EmitArgs = [Path | Error, Stats?];
```
paulmillr, line 62: Is that OK for backwards compat?

src/index.ts
```diff
@@ -567,8 +580,8 @@ export class FSWatcher extends EventEmitter {
   }
 
   emitWithAll(event: EventName, args: EmitArgs) {
-    this.emit(...args);
-    if (event !== EV.ERROR) this.emit(EV.ALL, ...args);
+    this.emit(event, ...args);
```
paulmillr, line 583: Huh?

paulmillr (changes requested):

src/index.ts
43081j: `args` used to contain the event name too and we basically didn't make use of `event` even though we passed it

I simplified so we use the event name and the args separately

We only had them combined for storing in the pending unlinks map. Everywhere else, we already had the name separately or didn't need it

src/index.ts
43081j: Sorry I forgot to reply to this. We do export it so it's technically a breaking change to the types. But as far as I'm aware, it's not part of our "public API". 

I couldn't find any usages of it in the wild. So I think it's really just an internal type exported for completeness

43081j:
cc @paulmillr 

i think we just need to cut a release of main, as we haven't done one in a while and have since merged a few things

mattjohnsonpint:
FYI, this was a breaking change for us.
https://github.com/hypermodeinc/modus/actions/runs/12397272157/job/34607246349

Fixed with https://github.com/hypermodeinc/modus/pull/655/commits/680f42280e2231d1963671e06c61fcd88641919b

paulmillr:
Why is the first argument error now @43081j ?

mattjohnsonpint:
For me, it was `string` and now it's `string | Error`.  I'm not sure in what circumstance an error would occur but the event would still be raised?

43081j:
it was always `string | Error` under the hood but the event handlers weren't strongly typed, so it was never enforced

this is because the generic code for emitting events can sometimes pass an error, and `unlink` doesn't have an explicit signature so falls back to that

we should probably just add `unlink` to the explicit ones here:

https://github.com/paulmillr/chokidar/blob/69c115a3684a5e72e0ac68aa0a57c204f081132a/src/index.ts#L301-L306

as something like:

```ts
[EV.UNLINK]: [string]
```

there may be other handlers in the same situation.

a quick scan around tells me `unlink` and `unlink_dir` will _always_ be a path (never an error)

meanwhile, it seems like `add` and `change` can be an `Error` (maybe others too)

43081j:
i've opened #1397 to at least fix the unlink events

there may be other events but we can look into that separately

mattjohnsonpint:
Makes sense.  Any suggestions for how to handle such errors? I think in my case, ignoring them is fine. Maybe some specific examples of the kinds of errors that could be passed would be helpful to others. Thanks.

paulmillr:
>meanwhile, it seems like add and change can be an Error (maybe others too)

How so? It was my assumption from day one it's always `path string`.

43081j:
i think its mostly in situations where `awaitWriteFinish` is set. if while polling the file, it vanishes or a `stat` fails, that error will be passed to the callback

i suspect you can mostly ignore them if that's the case (i.e. just skip your handler if it is an error)

paulmillr:
Can you produce a test example?

43081j:
its difficult to rep;roduce, but its basically here:

https://github.com/paulmillr/chokidar/blob/69c115a3684a5e72e0ac68aa0a57c204f081132a/src/index.ts#L637-L651

this gets called over time (polling) as it waits for the file size to settle

it is checking the file size each time via a stat call. so if one of those calls fails, or the file has been removed, it will pass an error back here:

https://github.com/paulmillr/chokidar/blob/69c115a3684a5e72e0ac68aa0a57c204f081132a/src/index.ts#L770-L773

43081j:
this `Path | Error` stuff was here before but it seems wrong now that i look at it

ill chop the code up a bit and see where i get to

43081j:
@paulmillr i've updated #1397 to try fix this, please take a look if you can

benmccann:
This was a breaking change for us too. I'm guessing changing `EventEmitter<FSWatcherEventMap>` to `EventEmitter`

The line `const ready = new Promise((resolve) => watcher.on('ready', resolve))` now causes:
```
src/index.js:81:61 - error TS2345: Argument of type '(value: any) => void' is not assignable to parameter of type '() => void'.
  Target signature provides too few arguments. Expected 1 or more, but got 0.

81  const ready = new Promise((resolve) => watcher.on('ready', resolve));
```

(tracking in https://github.com/sveltejs/kit/issues/13380)

43081j:
It's not wrong 

That should be a `Promise<void>` right? Would it then work? 

I understand it has broken code that worked with a previous 4.x release. But this is a difficult one because the types are actually right here, which suggests it was a bug fix rather than a break. 

So I'm not sure what to do here. To weaken the types would in this way would make them incorrect

benmccann:
We use JS, so the code can't change. But we can fix it by adding some JSDoc. I don't think chokidar needs to change anything, but thought I'd at least mention it here so that others can follow our fix if they run into the same thing. We've got a PR to address it on our side: https://github.com/sveltejs/kit/pull/13381

PR #1384, opened 2024-11-05
Ignore helix temp files

aster-void:
- [x] checked all tests pass

related issue (in Helix repo): https://github.com/helix-editor/helix/issues/11998

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -75,7 +75,7 @@ const TWO_DOTS = '..';
 const STRING_TYPE = 'string';
 const BACK_SLASH_RE = /\\/g;
 const DOUBLE_SLASH_RE = /\/\//;
-const DOT_RE = /\..*\.(sw[px])$|~$|\.subl.*\.tmp/;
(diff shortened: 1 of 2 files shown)


paulmillr:
you can use anything you'd like in `ignored`. There is no need to edit built-ins.

aster-void:
I see other editor's temp files ignored here, why shouldn't helix (a file editor fyi) be one of them? I think this is the most appropriate place to patch this.

paulmillr:
Brief google says it's not exclusive to helix https://file.org/extension/bck

Again, you can adjust this in your app. It doesn't matter what's the default configuration is.

aster-void:
neither as swp. https://file.org/extension/swp

Anyways, I'll open a patch request for vite. thanks anyways.

43081j:
It is only used for the atomic writes feature FYI. Since if you enable that, you want to tell chokidar to detect temp editor files that happen mid write (like swap files) 

That's why we wouldn't usually change it

PR #1397, opened 2024-12-18
fix: separate error emit args from regular emit args

43081j:
Separates the error event arguments from the regular event arguments.

Before, we had a merged tuple of `[Path | Error, Stats?]` to account for
the fact that some events can be `[Path, Stats?]` and some can be
`[Error, Stats?]`.

However, this results in incorrect types since nothing will ever pass
`Path | Error` as a type (only one or the other).

This separates them and uses a union instead, such that event handlers
other than `error` only ever have a `Path`.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -303,6 +303,8 @@ export interface FSWatcherKnownEventMap {
   [EV.RAW]: Parameters<WatchHandlers['rawEmitter']>;
   [EV.ERROR]: Parameters<WatchHandlers['errHandler']>;
   [EV.ALL]: [EventName, ...EmitArgs];
+  [EV.UNLINK]: [path: string];
+  [EV.UNLINK_DIR]: [path: string];
 }
 
 export type FSWatcherEventMap = FSWatcherKnownEventMap & {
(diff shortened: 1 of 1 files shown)


JounQin:
I got the following error after upgrading to v4.0.3

```sh
src/cli/index.ts:87:20 - error TS2576: Property 'on' does not exist on type 'FSWatcher'. Did you mean to access the static member 'FSWatcher.on' instead?

87         cliWatcher.on('all', async (eventName, filepath) => {
                      ~~

src/cli/index.ts:87:37 - error TS7006: Parameter 'eventName' implicitly has an 'any' type.

87         cliWatcher.on('all', async (eventName, filepath) => {
                                       ~~~~~~~~~

src/cli/index.ts:87:48 - error TS7006: Parameter 'filepath' implicitly has an 'any' type.

87         cliWatcher.on('all', async (eventName, filepath) => {
                                                  ~~~~~~~~
```

43081j:
we haven't changed anything around that as far as i can tell. are you sure it didn't already error in 4.0.2 with `@types/node@18.18`?

and you just coincidentally upgraded both at the same time?

otherwise seems pretty strange as this PR didn't touch what node types we use

PR #1403, opened 2025-01-19
test: switch to node:fs/promises

43081j:
Switches to `node:fs/promises` in tests

This also caught a couple of places where we were not awaiting FS operations, leading to flaky tests. Those are awaited now and should improve test runs.


43081j:
i didn't realise main has a bunch of conflicting changes, will figure it out

test.mjs
```diff
@@ -205,7 +197,7 @@ const runTests = (baseopts) => {
       const spy = sinon.spy(function addSpy() {});
       watcher.on(EV.ADD, spy);
       await delay();
-      await write(testPath, dateNow());
+      await fsp.writeFile(testPath, dateNow());
```
paulmillr, line 200: I liked the alias more. Can you assign write = fsp.writeFile? Same for rm, read.

43081j: Sure can do 👍

PR #1404, opened 2025-01-19
test: strongly typed tests

43081j:
I'm leaving this as a draft for now just to show you what I was thinking @paulmillr

Basically, this moves the tests into `src/` and converts them to typescript. I then ignored them from published files, etc.

All seems to work but these things had to change or need discussing (if we ever want this):

- we need `node16` resolution/module so `*.mts` is treated as ESM and `import.meta` exists
  - i manually diffed the build output before/after this change and it is identical
- we have a test which tries to mock `fs.Stat`, but this isn't possible in ESM
  - i've commented on the relevant test

### DIFF
--- a/src/index.test.mts
+++ b/src/index.test.mts
@@ -1,19 +1,19 @@
-import fs from 'node:fs'; // fs.stat is mocked below, so can't import INDIVIDUAL methods
+import * as fs from 'node:fs'; // fs.stat is mocked below, so can't import INDIVIDUAL methods
 import * as fsp from 'node:fs/promises';
 import { writeFile as write, readFile as read, rm } from 'node:fs/promises';
-import sysPath from 'node:path';
+import * as sysPath from 'node:path';
 import { describe, it, beforeEach, afterEach } from 'micro-should';
 import { fileURLToPath, pathToFileURL, URL } from 'node:url';
 import { promisify } from 'node:util';
 import { exec as cexec } from 'node:child_process';
 import { tmpdir } from 'node:os';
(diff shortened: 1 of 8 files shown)


src/index.test.mts
```diff
@@ -1618,36 +1619,32 @@ const runTests = (baseopts) => {
...
+        let statStub: sinon.SinonSpy;
+
         beforeEach(() => {
           options.awaitWriteFinish = { pollInterval: 50, stabilityThreshold: 50 };
           options.ignoreInitial = true;
 
           // Stub fs.stat() to take a while to return.
-          sinon.stub(fs, 'stat').callsFake((path, cb) => {
-            _realStat(path, w(cb, 250));
+          statStub = sinon.stub(fs, 'stat').callsFake(async (path, cb) => {
```
43081j, line 1645: this test can't have ever done what we thought since we moved to ESM here (i.e. since this became `test.mjs`)

in ESM you can't stub out a module's exports like this. which means it won't have been delaying the `fs.stat` calls

the test has been passing though... so maybe it just didn't actually test the thing it was trying to test? and passed because no such delay happened

paulmillr:
Thanks @43081j  -- looks great.

src/index.test.mts
paulmillr: we can remove it

paulmillr (approved):

43081j:
I'll remove the disabled test in a follow-up if you don't beat me to it 👍

PR #1407, opened 2025-01-23
test: remove fs mocked test

43081j:
Removes the skipped test since we can't really do it in ES modules anyway (as we can't mock ESM exports).


43081j:
@paulmillr there are a couple of other tests skipped. do you know why?

this one:

https://github.com/paulmillr/chokidar/blob/7b3e627739b0397dc5759882b6ec4442295228e2/src/index.test.mts#L674

and this one:

https://github.com/paulmillr/chokidar/blob/7b3e627739b0397dc5759882b6ec4442295228e2/src/index.test.mts#L1770

do you know if/how we can re-enable them?

paulmillr:
they've been failing forever I think. Hard to remember.

PR #1408, opened 2025-01-24
test: switch from `should` style to `expect`

43081j:
Switches from using `should` assertions (e.g. `foo.should.equal(303)`) to `expect` assertions (e.g. `expect(foo).to.equal(303)`).

This is a stepping stone to moving off `sinon` and possibly `chai` in future.

FYI this was automated using an ast-grep codemod [found here](https://github.com/43081j/codemods/blob/main/codemods/chai-should-to-expect.yml) (which itself has tests)

### DIFF
--- a/src/index.test.mts
+++ b/src/index.test.mts
@@ -19,7 +19,6 @@ const TEST_TIMEOUT = 32000; // ms
 
 const { expect } = chai;
 chai.use(sinonChai);
-chai.should();
 
 const exec = promisify(cexec);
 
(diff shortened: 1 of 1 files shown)


43081j:
just found why one of the tests was flaky too!

it was waiting for 3 events instead of 5... so fixed it while i was in there

paulmillr:
Why don’t we just move to assert? Or maybe assert has expect inside?

paulmillr:
I am not a huge fan of assert, but at least it is everywhere

43081j:
We are moving to assert, I'm just not doing it in one massive PR as it'll be difficult to review

1. Move from `should` to `expect` so all assertions are consistent (we use a mixture right now)
2. Move from `expect` to `assert`
3. Move from `sinon` to `tinyspy`

Roughly this is my plan

43081j:
i've opened #1409 which we can aim to merge once this PR lands

then we will have moved to `node:assert` fully

PR #1413, opened 2025-01-27
test: migrate to tinyspy

43081j:
Migrates from `sinon` to the much smaller `tinyspy`.

To avoid repeating logic, some helpers have also been added (e.g. `calledWith`).

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,10 +33,9 @@
   "devDependencies": {
     "@paulmillr/jsbt": "0.3.1",
     "@types/node": "22.10.10",
-    "@types/sinon": "17.0.3",
     "micro-should": "0.5.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -33,10 +33,9 @@
   "devDependencies": {
     "@paulmillr/jsbt": "0.3.1",
     "@types/node": "22.10.10",
-    "@types/sinon": "17.0.3",
     "micro-should": "0.5.1",
     "prettier": "3.1.1",
-    "sinon": "12.0.1",
+    "tinyspy": "^3.0.2",
```
paulmillr, line 38: Please do not use version range for supply chain security reasons

paulmillr: already done

PR #1430, opened 2025-06-14
Migrate to ESM only

### LINKED ISSUE #1429: Release pure esm v5 with node v20+ requirement
- chokidar v4 supports node v14+
- LTS nodejs supported as per june 2025 are v20, v22, v24 (no v14, v16, v18)
- All current LTS nodes support importing ESM modules from common.js
- The feature is not present in v18

I suggest we release pure-ESM v5 without other changes. It would be easier to maintain and more modern. v4 could still be around for the time being - it's still better than v3.

43081j:
Fixes #1429 

cc @paulmillr 

i ran it through [arethetypeswrong](https://github.com/arethetypeswrong/arethetypeswrong.github.io) and all seems to be good


paulmillr:
1. “Files” control which files are in package, not “exports”. I think it doesn’t matter if *handler files are exposed.
2. In the past there were some issues with ts not picking it up even while being in same root dir. maybe it’s fixed now
3. jsbt has erasable syntax tsconfig setting, which we inherit — but useful to test it manually just in case

43081j:
1. Files controls what we ship, exports controls what you can import. But again I don't mind either way since we're exporting every js file pretty much
2. It should be fine now 👍 
3. I can add a test when I get chance

43081j:
we can't actually do the typescript test since chokidar itself can't be loaded that way

all of our imports correctly reference `.js` files right now, since that's what will exist at run-time

if we wanted to run chokidar itself from typescript sources, we'd need to configure typescript to use `.ts` extension in imports (i.e. transform it from `.ts` to `.js` at build time. since the run-time js still needs `.js` imports but the sources need `.ts` imports)

though im not sure any of this matters to us. anyone can import chokidar fine right now in their own typescript executed directly through node. the only people who would want to run chokidar itself on the command line are probably us

paulmillr:
> transform import paths from .ts to .js at build time

that’s what I do in all noble repos. Jsbt is already prepared for it.

43081j:
Where would we be executing chokidar itself through node?

Won't it always be imported? 

I know we can flip the flag in tsconfig. Just want to understand where we'd use it though

Either way we probably should tackle it in a separate issue/pr

PR #1432, opened 2025-06-15
ci: use Nodejs 24

btea:


btea:
I think that's more reasonable. 👍

PR #1433, opened 2025-06-15
chore: consolidate CI stages

43081j:


paulmillr:
It would be better to reuse jsbt actions https://github.com/paulmillr/noble-hashes/tree/main/.github/workflows

there is test-js and test-ts, difference is in: one tests compiled js code, other raw ts. (We should prob use test-js)

43081j:
because we need to use windows, linux and mac here but jsbt only uses one OS at a time

paulmillr:
Hmmm that's my screwup; i've thought matrix in jsbt has windows and mac as well.

Well...

43081j:
this is all sorted now but the extra runs are showing up our known-to-be-flaky test (the nine directories one, and maybe the thirtythree test)

would be nice to figure out why it is a flaky test

PR #1441, opened 2025-11-10
Simplification of toUnix and normalizePath

Pjb518:
Simplifies the implementation of `toUnix` and `normalizePath`. Adds a negative lookbehind for the `DOUBLE_SLASH_RE` constant. Removes the now unused `DOUBLE_SLASH` constant.

### DIFF
--- a/src/handler.ts
+++ b/src/handler.ts
@@ -531,7 +531,7 @@ export class NodeFsHandler {
     directory: string,
     initialAdd: boolean,
     wh: WatchHelper,
-    target: Path,
+    target: Path | undefined,
     dir: Path,
     depth: number,
     throttler: Throttler
@@ -539,7 +539,8 @@ export class NodeFsHandler {
     // Normalize the directory name on Windows
     directory = sp.join(directory, '');
 
-    throttler = this.fsw._throttle('readdir', directory, 1000) as Throttler;
+    const throttleKey = target ? `${directory}:${target}` : directory;
+    throttler = this.fsw._throttle('readdir', throttleKey, 1000) as Throttler;
     if (!throttler) return;
 
     const previous = this.fsw._getWatchedDir(wh.path);
@@ -632,7 +633,7 @@ export class NodeFsHandler {
     stats: Stats,
     initialAdd: boolean,
     depth: number,
-    target: string,
+    target: string | undefined,
     wh: WatchHelper,
     realpath: string
   ): Promise<(() => void) | undefined> {
@@ -713,7 +714,7 @@ export class NodeFsHandler {
           stats,
           initialAdd,
           depth,
-          target!,
+          target,
           wh,
           targetPath
         );
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -195,7 +195,6 @@ function waitForEvents(watcher: chokidar.FSWatcher, count: number) {
 
 const runTests = (baseopts: chokidar.ChokidarOptions) => {
   let macosFswatch = isMacos && !baseopts.usePolling;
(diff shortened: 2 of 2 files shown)


Pjb518:
> you're basically suggesting runtime type checking, which i'm not sure i agree with.
> 
> we statically type this to already accept the right shape, so this `typeof` check is merely to narrow between the strong types we know we will receive: `string | RegExp | MatchFunction | MatcherObject`.
> 
> if someone passes an object which doesn't satisfy this type definition, i don't think we need to care, that's their own fault.

My proposed check differs from the current one only in that we check for a `path` attribute rather than ensuring the object is not a regular expression. The goal is less about runtime type checking and more about more effectively moving junk matchers into the "this is garbage" `() => false` path.

Checking for a string `path` property rules out regex just as well as checking `instanceof`, but it also catches other unusable object types. In practice, that means running the following chain far less for objects we could easily verify up front are not going to work:
```js
if (matcher.path === string) return true;
if (matcher.recursive) {
  const relative = sp.relative(matcher.path, string);
  if (!relative) {
    return false;
  }
  return !relative.startsWith('..') && !sp.isAbsolute(relative);
}
```

43081j:
```ts
if (matcher.path === string) return true;
if (matcher.recursive) {
  const relative = sp.relative(matcher.path, string);
  if (!relative) {
    return false;
  }
  return !relative.startsWith('..') && !sp.isAbsolute(relative);
}
```

this only happens as part of the test function of a matcher _object_ (not a RegExp)

```ts
  if (matcher instanceof RegExp) return (string) => matcher.test(string);
  if (typeof matcher === 'object' && matcher !== null) {
    return (string) => {
      if (matcher.path === string) return true;
      if (matcher.recursive) {
        const relative = sp.relative(matcher.path, string);
        if (!relative) {
          return false;
        }
        return !relative.startsWith('..') && !sp.isAbsolute(relative);
      }
      return false;
    };
  }
```

everything inside the anonymous function assumes `matcher` is a matcher object.

> In practice, that means running the following chain far less for objects we could easily verify up front are not going to work:

so i don't understand this part. this function is only called for matcher objects, what am i missing? 🤔 

> Checking for a string path property rules out regex just as well as checking instanceof, but it also catches other unusable object types

we have strongly typed what our function accepts. if the user passes nonsense into it, i think its their own fault. 

i feel like you're trying to guard against invalid inputs from users. but im saying we don't need to care because we've strongly defined what we accept and its your own fault if you pass something that isn't that. this is already correctly narrowing for the types we accept. we don't need to check if `path` exists because, if its not a RegExp, it must be a valid matcher object as per our types.

43081j:
> why not check its the type we actually need) tells us that the object is not null, but it also tells us whether the object has the absolute bare minimum properties to potentially generate a useful matcher function.

because we told you up front what types the function accepts. if you pass one that doesn't conform to those types its your own fault and it'll probably cause uncaught exceptions.

> If that's true, then why do we need to check for null, and why do we need that final default case where we provide () => false; as the matcher function where we couldn't trigger any of the conditions?

to satisfy TypeScript. its a branch nobody should ever hit but needs to exist to satisfy the type system.

> The thing is, we just say that the options we're getting from the user are Matchers, but I don't think we actually check that anywhere.

we strongly type the options to define that matchers are of type `Matcher`, statically. we don't do any runtime type checking because its your own fault if you get around the static types and will cause uncaught exceptions. the checks we do are just to narrow the union correctly, not to do any kind of validation.

43081j:
@Pjb518 can you catch up from main?

we merged a bug fix that might sort the tests too. the bun tests will fail but that is expected

PR #1442, opened 2025-11-11
test: wait for explicit calls in directory test

### LINKED ISSUE #1437: Events are dropped when watching multiple files from one FSWatcher
## Description
When watching more than one file (non-polling), atomic writes on any files other than the first one that's changed are emitted as an `unlink` event, and then no more events are emitted for that file. The file that was written first continues to properly emit all events.

## Versions:
 - Chokidar: `4.0.3`
 - Node: `v22.14.0`
 - OS: `NixOS 25.05.802491.7c43f080a7f2 (Warbler)`

## Reproducing

```typescript
import { watch } from "chokidar";

watch(["a", "b", "c"], {
    persistent: true,
    atomic: 500, // delay doesn't seem to affect bug behavior
}).on("all", (...a) => console.log(a));
```
*It's assumed that all three files (`a`, `b`, and `c`) already exist*

- Update file `a` with an atomic write (I use neovim). This will emit a `change` event.
- Update files `b` and `c`. Both of these will emit an `unlink` event after 500ms (or whatever delay was set)

## Expected Behavior

All file updates should emit a change event

## Context and Thoughts

I've reproduced this issue on NixOS and Ubuntu on WSL, so I don't think it's system specific. Also, watching each file with a completely separate `FSWatcher` works just fine, but that's effectively what the single watcher should be doing under the hood anyway so I can't imagine why. I imagine the problem is just a minor logic bug, so I'll look at the source myself to see if I can figure it out. Can't be much harder than building a workaround.... right?

43081j:
This hopefully fixes the flakiness since we explicitly wait for each
step rather than hoping the delay was right.

Fixes #1437 too

### DIFF
--- a/src/handler.ts
+++ b/src/handler.ts
@@ -531,7 +531,7 @@ export class NodeFsHandler {
     directory: string,
     initialAdd: boolean,
     wh: WatchHelper,
-    target: Path,
+    target: Path | undefined,
     dir: Path,
     depth: number,
     throttler: Throttler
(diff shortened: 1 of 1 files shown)


43081j:
well i can reproduce the bun problem now so will dig into it before i merge this or #1441 

it feels like a bun bug but not sure yet

PR #1449, opened 2026-01-10
Watch for return of unlinked files

artygus:
Problem manifests itself when one of the watched files gets removed and added back later on, chokidar emits unlink event and never adds file back. I hit this problem with tsup that [uses chokidar](https://github.com/egoist/tsup/blob/b906f86102c02f1ef66ebd4a3f7ff50bdc07af65/src/index.ts#L406-L409) under the hood, I guess it's because vim actually replaces edited file instead of directly writing to the same inode.

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -702,8 +702,7 @@ const runTests = (baseopts: chokidar.ChokidarOptions) => {
       equal(getCallsWith(spy, [EV.CHANGE, testPath]).length, 3);
     });
 
-    // PR 682 is failing.
-    describe.skip('Skipping gh-682: should detect unlink', () => {
+    describe('should detect unlink', () => {
       it('should detect unlink while watching a non-existent second file in another directory', async () => {
         const testPath = dpath('unlink.txt');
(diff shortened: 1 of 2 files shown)


43081j:
this is a really difficult one to review 😅 

the logic you changed is 12+ years old. and we certainly have tests which assert removed files are detected when reintroduced.

e.g. we have this test:

```ts
    it('should emit `add`, not `change`, when previously deleted file is re-added', async () => {
```

which does work before this PR.

so can you help me understand which specific remove/add case this is fixing? i'm sure you have fixed something, but it isn't "return of unlinked files" because there are already tests passing for that. so it must be a _subset_ of it, something more edge case.

understanding that will help me review

artygus:
Hi @43081j, half of the tests I've uncommented were failing and it helped in my specific case which is tsup. I understand it's hard to review, but code of chokidar is super hard to follow too, I can't give a clear explanation on why it helps and in which cases it won't.

Thanks for taking a look, but I'm gonna close down the PR, since I've migrated away from tsup, my new building system doesn't use chokidar.

PR #1456, opened 2026-04-10
fix #1455 - event throttling discards events

### LINKED ISSUE #1455: Event throttling discards updates
We are using chokidar for a custom build pipeline and if we switch branches, it sometimes happens that a file is changed multiple times within 50ms. Because of the throttling introduced in https://github.com/paulmillr/chokidar/commit/b4e9430ef8932f8d6ef4ab2468b2bd1e1bbb5d6d we only get the first of these changes and no final event, and thus our build output is wrong, because we act on outdated information and do not get a later signal again.
I would expect a final event when the throttling interval has passed.

Removing the block: https://github.com/paulmillr/chokidar/blob/main/src/index.ts#L657 fixed it for us.

Could you please clarify what this codepath is actually used for?

**Describe the bug**

 - Chokidar version: 5.0.0
 - Node version: v24.14.1
 - OS version: MacOS 26.4

**To Reproduce:**
```diff
diff --git a/src/index.test.ts b/src/index.test.ts
index 453e1c6..c608a71 100644
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -432,6 +432,18 @@ const runTests = (baseopts: chokidar.ChokidarOptions) => {
       ok(rawSpy.called);
       equal(spy.callCount, 1);
     });
+    it('should emit two `change` events for two writes within 50ms', async () => {
+      const testPath = dpath('change.txt');
+      const spy = createSpy<EmitArgs, void>(function quickChangeSpy() {});
+      watcher.on(EV.CHANGE, spy);
+
+      await write(testPath, time());
+      await delay(25);
+      await write(testPath, time());
+
+      await waitFor([[spy, 2, [testPath]]]);
+      ok(getCallsWith(spy, [testPath]).length >= 2);
+    });
     it('should emit `unlink` event when file was removed', async () => {
       const testPath = dpath('unlink.txt');
       const spy = createSpy<EmitArgs, void>(function unlinkSpy() {});
```

**Expected behavior**
I would expect that after the throttle period has passed, at least one more change event would be triggered, if any change events occureed during the throttle period.

adrianjost:
See issue for more details: https://github.com/paulmillr/chokidar/issues/1455

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -432,6 +432,36 @@ const runTests = (baseopts: chokidar.ChokidarOptions) => {
       ok(rawSpy.called);
       equal(spy.callCount, 1);
     });
+    it('should emit two `change` events for two writes within 50ms', async () => {
+      const testPath = dpath('change.txt');
+      const spy = createSpy<EmitArgs, void>(function quickChangeSpy() {});
+      watcher.on(EV.CHANGE, spy);
+
+      await write(testPath, time());
+      await delay(10);
+      await write(testPath, time());
+
+      await waitFor([[spy, 2, [testPath]]]);
+      ok(getCallsWith(spy, [testPath]).length >= 2);
+    });
+    it('should not emit `change` after `unlink` when changes were throttled', async () => {
+      const testPath = dpath('change.txt');
+      const changeSpy = createSpy<EmitArgs, void>(function throttledChangeSpy() {});
+      const unlinkSpy = createSpy<EmitArgs, void>(function throttledUnlinkSpy() {});
+      watcher.on(EV.CHANGE, changeSpy).on(EV.UNLINK, unlinkSpy);
+
+      await write(testPath, time());
+      await delay(10);
+      await write(testPath, time());
+      await delay(10);
+      await unlink(testPath);
+
+      await waitFor([[unlinkSpy, 1, [testPath]]]);
+      const changeCallsAtUnlink = getCallsWith(changeSpy, [testPath]).length;
+
+      await delay(120);
+      equal(getCallsWith(changeSpy, [testPath]).length, changeCallsAtUnlink);
+    });
     it('should emit `unlink` event when file was removed', async () => {
       const testPath = dpath('unlink.txt');
       const spy = createSpy<EmitArgs, void>(function unlinkSpy() {});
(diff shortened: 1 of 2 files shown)


adrianjost:
Interestingly the tests run successful in a GitHub CodeSpace with node v24.11.1, but apparently not on Node v22. 🫤

43081j:
im not sure this is the right change.

your specific problem is that you want the last change event, not that you want _all_ events. this throttle exists for good reason - because different OSes and different engines can emit multiple change events for the same file in quick succession.

if we remove this throttle, its pretty much a breaking change because we'll start emitting a lot more events for all users.

solving your specific problem at the cost of changing behaviour for everyone isn't the right move.

i think the right fix, if any, would be to keep the throttle and emit the last event rather than the first. but this needs a lot of thought because that is a rather fundamental change. ill ponder on it and come back here some time

adrianjost:
I appreciate the perspective, but I’m not entirely convinced that adding more logic on top of the throttle is the cleanest path forward.

My concern is that we’d be patching a behavior that is currently 11 years old, undocumented in [commit history](https://github.com/paulmillr/chokidar/commit/b4e9430ef8932f8d6ef4ab2468b2bd1e1bbb5d6d), and lacks a reliable test suite. In my view, simplifying the behavior to something predictable is often safer than maintaining a complex throttle. I also noticed the current CI failures appear to be flaky (It only failed 1/10 runs in my Node v22 Codespace), which suggests the existing state might already be less stable than we'd like.

That said, I don't want to be pushy or create more work for you. We’ve actually switched to fs.watch for our specific requirements, so there is no pressure on this PR. Feel free to use these findings as a reference or simply close this out if you'd prefer to keep the architecture as-is. Thanks for all the work you've already put into this library!

43081j:
the tests are failing because 2 events were seen rather than 1. that seems too much of a coincidence given the change is to make it emit more events 😅 

outside of this, the windows tests are flaky in general just because windows events are less reliable than linux ones in our experience.

the concern is that changing this means more events will be emitted for everyone. of course _simpler_ logic is easier to maintain, but that doesn't change the fact you're making a globally breaking change all users will hit (since this is the hot path).

so it just needs some thought before we can do it. meanwhile if you could be sure the test failure is unrelated to your changes, that'd be helpful

adrianjost:
I have managed to reproduce the failing CI now, and it seems indeed caused by my changes. Surprisingly not on the Linux based Codespace, only on MacOS, but I guess that's just one of the many file system quirks. I will see if I can fix this and come back to you.

src/index.test.ts
```diff
@@ -1461,17 +1491,20 @@ const runTests = (baseopts: chokidar.ChokidarOptions) => {
...
 
         await unlink(dpath('unlink.txt'));
         await write(dpath('change.txt'), time());
-        await Promise.all([watcherEvents, watcher2Events]);
+        await waitFor([
+          [spy1, 1, [EV.CHANGE, 'change.txt']],
+          [spy1, 1, [EV.UNLINK, 'unlink.txt']],
+          [spy2, 1, [EV.CHANGE, sp.join('..', 'change.txt')]],
+          [spy2, 1, [EV.UNLINK, sp.join('..', 'unlink.txt')]],
+        ]);
```
adrianjost, line 1507: This test needed adjustment, because the `waitForEvents` counter was not working reliable anymore. Depending on how fast the file changes before were made, there were either 5 or 6 events due to throttling. I've switched to explicitly waiting for the events we are expecting below.

adrianjost:
@43081j I think I now found a way to keep the throttling but emit a final change event after the changes have settled. Only a single test needed some minor setup adjustment. Let me know what you think and if I misunderstood something.

PR #1457, opened 2026-04-27
chore: bump dev dependencies

43081j:
This updates typescript and a few other dev dependencies.

The type checker got smarter so one of the test functions no longer type
checks correctly thanks to spying on a complex union. So I have used an
`as never` cast.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,12 +12,12 @@
         "readdirp": "^5.0.0"
       },
       "devDependencies": {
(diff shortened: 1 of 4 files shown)


paulmillr:
Could you please lock down dev dep versions? @43081j

43081j:
Ah yeah sorry i forgot we had them pinned. I'll do a follow up with them pinned again 👍

PR #1458, opened 2026-04-29
fix: default cwd to process.cwd() for relative ignore paths

### LINKED ISSUE #1436: Can't ignore relative paths
Absolute and regex (`**`) paths included in the config options are properly ignored, but relative paths never work. e.g.:

```ts
ignored: ['build', 'tests'']
```

Won't ignore those folders and will continue to watch them.

The issue is that relative paths are combined with the `cwd` but the `cwd` defaults to an empty string. Using the empty string means relative paths will never work, unless you happen to be in the root directory. You can see this in several places in `index.ts`, such as the `normalizeIgnored` function: https://github.com/paulmillr/chokidar/blob/main/src/index.ts#L182

Possible solution: use `process.cwd()` as the default instead. I can create a pull request if this is acceptable.

This is happening on all platforms with the most recent published version, 4.0.3, though this issue has existed since at least October, 2023: https://stackoverflow.com/questions/77355876/vite-not-respecting-server-watch-ignored-option-laravel/79724226

I would hazard to guess that there are a lot of projects out there thinking they're ignoring relative folders, but actually aren't because of this bug. They just won't know it until they hit the file watcher limit.

abhu85:
## Summary
Default `cwd` option to `process.cwd()` instead of empty string to fix relative path matching in `ignored`.

## Problem
When `cwd` defaults to `""`, relative paths in the `ignored` option are joined with an empty string, producing paths that never match actual file paths. This means `ignored: 'node_modules'` silently fails to ignore anything unless `cwd` is explicitly set.

## Solution
Default `cwd` to `process.cwd()` when not provided in `normalizeIgnored`. This matches the intuitive behavior and was confirmed as the correct approach by collaborator @43081j.

## Test Plan
- [x] `ignored: 'relative-dir'` works without explicit `cwd`
- [x] Explicit `cwd` still works as before
- [x] Existing test suite passes (207 tests, 0 failures)

Fixes #1436

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -1278,6 +1278,29 @@ const runTests = (baseopts: chokidar.ChokidarOptions) => {
         equal(calledWith(spy, [EV.ADD, testFile]), false);
         equal(calledWith(spy, [EV.CHANGE, testFile]), false);
       });
+      it('should ignore relative paths without explicit cwd', async () => {
+        const testDir = dpath('ignored-dir');
+        const testFile = sp.join(testDir, 'add.txt');
+        options.ignored = 'ignored-dir';
+        await mkdir(testDir);
+        await write(testFile, 'b');
+        const prevCwd = process.cwd();
+        process.chdir(currentDir);
+        try {
+          const watcher = cwatch(currentDir, options);
+          const spy = await aspy(watcher, EV.ALL);
+
+          await delay();
+          await write(testFile, time());
+
+          await delay(300);
+          equal(calledWith(spy, [EV.ADD_DIR, testDir]), false);
+          equal(calledWith(spy, [EV.ADD, testFile]), false);
+          equal(calledWith(spy, [EV.CHANGE, testFile]), false);
+        } finally {
+          process.chdir(prevCwd);
+        }
+      });
       it('should allow regex/fn ignores', async () => {
         options.cwd = currentDir;
         options.ignored = /add/;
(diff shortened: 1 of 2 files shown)


src/index.test.ts
```diff
@@ -1278,6 +1278,29 @@ const runTests = (baseopts: chokidar.ChokidarOptions) => {
...
+          await write(testFile, time());
+
+          await delay(300);
+          equal(calledWith(spy, [EV.ADD_DIR, testDir]), false);
+          equal(calledWith(spy, [EV.ADD, testFile]), false);
+          equal(calledWith(spy, [EV.CHANGE, testFile]), false);
+        } finally {
+          process.chdir(prevCwd);
+        }
+      });
```
43081j, line 1303: we should add a test for when cwd is set too:

```suggestion
      });
      it('should ignore contents of relative dir with cwd set', async () => {
        const testDir = dpath('subdir');
        const testFile = sp.join(testDir, 'add.txt');
        options.ignored = 'subdir';
        options.cwd = currentDir;
        await mkdir(testDir);
        await write(testFile, 'b');
        const watcher = cwatch(currentDir, options);
        const spy = await aspy(watcher, EV.ALL);

        await delay();
        await write(testFile, time());

        await delay(300);
        equal(calledWith(spy, [EV.ADD_DIR, testDir]), false);
        equal(calledWith(spy, [EV.ADD, testFile]), false);
        equal(calledWith(spy, [EV.CHANGE, testFile]), false);
      });
```

abhu85: Added in fb96052. Both tests pass in polling and non-polling modes (209 total).

PR #1459, opened 2026-04-30
feat: support `recursive: true` (DRAFT/WIP)

43081j:
This introduces a new `recursive: boolean` option to the watcher.

Once set to `true`, the following will happen:

- When `_addToNodeFs` is called, we pass through to a new `_handleDirRecursive` codepath and hit the existing `_handleDir` otherwise
  - If polling is enabled, this is skipped
  - If we're watching a specific target (symlinks or single files), this is skipped
- The new `handleDirRecursive` is called with the dir we want to watch
- The directory is emitted as `ADD_DIR` if `ignoreInitial` is off
- The directory is noted (but no additional watcher added)
- We then do a one-off traversal of the directory to add any existing items
- Finally, we add or update a recursive `fs.watch` listener

In terms of listeners, the main difference is that a recursive listener deals directly with the _emitted_ path. A non-recursive listener deals with the _file being watched_.

Beyond that, everything should be roughly the same:
- If the item doesn't exist somehow, remove it from the watcher
- If the item is a directory, emit `ADD_DIR`
- If the item is a file and was not previously tracked, emit `ADD`
- If the item is a file and was previously tracked, emit `CHANGE`

**To avoid huge reworks, I tried to keep this new codepath standalone/separate**

### DIFF
--- a/src/handler.ts
+++ b/src/handler.ts
@@ -43,6 +43,7 @@ const KEY_LISTENERS = 'listeners';
 const KEY_ERR = 'errHandlers';
 const KEY_RAW = 'rawEmitters';
 const HANDLER_KEYS = [KEY_LISTENERS, KEY_ERR, KEY_RAW];
+const BACKSLASH_PATTERN = /\\/g;
 
 // prettier-ignore
 const binaryExtensions = new Set([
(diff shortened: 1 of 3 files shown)


paulmillr:
Perf benchmarks would be useful to understand how this plays out.

Specifically:

1. Time to initially scan / watch large directories (10k, 100k, 400k, 800k, 1.2M files)
2. CPU load
3. CPU load post-initial scan

Also, cross-platform comparison would be even more useful.

paulmillr:
#1469

PR #1461, opened 2026-05-12
chore: set permissions on workflows

43081j:
Sets the permissions for the test workflow to be empty (i.e. no permissions).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,8 +4,8 @@ on:
     types: [created]
 jobs:
   release-js:
-    name: 'jsbt v0.5.1' # Should match commit below
-    uses: paulmillr/jsbt/.github/workflows/release.yml@a00dbc815760211c6f6986344b2da91512597810
+    name: 'jsbt v0.6.0' # Should match commit below
+    uses: paulmillr/jsbt/.github/workflows/release.yml@7c25cb0b5eadac32bfedbb3b60ca0bd4289791f1
     permissions:
       contents: read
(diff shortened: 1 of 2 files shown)


43081j:
bleh, @paulmillr i guess our CI is borked until this weird windows problem is resolved?

i saw you commented in libuv repo about it. any idea where its from?

paulmillr:
No idea

but likely not our fault.

paulmillr:
Ok we should disable node 26 until it's fixed

paulmillr:
@43081j what's the purpose of this PR's change? Could you detail me in? I'm keeping tabs on the tanstack malware story but can't understand how it's related.

43081j:
@paulmillr basically the default permissions of the overall workflow are whatever we define in the repo settings (which is either "read" or "read and write").

this just makes the default "no permissions", forcing us to set the right permissions per-step when needed.

paulmillr:
Ok and what would happen if the workflow has write permissions? What kind of stuff is possible?

43081j:
it means the GITHUB_TOKEN can be used to write to the repo for example.

in our case, we know our workflow isn't doing any of that. its more just a precaution so if we ever add new steps, we force ourselves to explicitly set the permissions on it

43081j:
@paulmillr i bumped to jsbt 0.6.0 here too and switched to `test-matrix` (used to be `test-ts-matrix`)

im going to merge FYI but CI is borked until libuv cuts a new release anyway sadly

PR #1462, opened 2026-05-12
fix: reintroduce graceful symlink infinite recursion

43081j:
The test for this originally used to assert that the watcher **did not
throw**.

Then in e487878 it was reversed to assert that the watcher **does
throw**.

It turns out the test passes because the OS throws an ELOOP error when
trying to stat a very deeply nested symlink. _Not_ because chokidar
threw an error, i.e. chokidar actually recurses into it very deeply
adding duplicate watchers until the OS throws.

**This change reverts the test to what it was originally, and changes the
gating to check a new set of seen realpaths rather than `_symlinkPaths`.**

Basically it seems:

- `_symlinkPaths` is a map of `symlinkPath -> realPath`
- This means the `!_symlinkPaths.has(realpath)` would only be hit if a
  symlink resolves to another symlink, or we already visited this exact
  symlink
- Changing it to use a separate map of visited directories means we can simply check if we resolve to a directory above ourselves that we already visited (i.e. a circular ref)

Doing a bunch of git research - it seems this back-and-forth was caused
by the fact we were using `_symlinkPaths` for both:

- Cycle detection (this logic)
- Unwatch cleanup (8f08914 tried to fix that but broke cycle detection)

### DIFF
--- a/src/handler.ts
+++ b/src/handler.ts
@@ -650,7 +650,8 @@ export class NodeFsHandler {
     let closer;
 
     const oDepth = this.fsw.options.depth;
-    if ((oDepth == null || depth <= oDepth) && !this.fsw._symlinkPaths.has(realpath)) {
+    const alreadyDescended = this.fsw._watched.has(realpath) && realpath !== dir;
+    if ((oDepth == null || depth <= oDepth) && !alreadyDescended) {
       if (!target) {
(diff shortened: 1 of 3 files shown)


43081j:
sorry just iterating a bit as it turns out its not as straightforward as i thought 😅 

edit; im pretty sure this change is good now but CI is being flaky...

PR #1467, opened 2026-07-06
fix: unwatch deleted watched directory on Windows

sapphi-red:
On windows, the watch directory was not unwatched when that directory is deleted.

`_closers` have the normalized paths stored (done by `normalizePathToUnix`). But `_closeFile` received a non-normalized path and thus the `this._closers.get` look up failed.

This caused https://github.com/vitejs/vite/issues/22672. Note that while Vite uses v3, the same code path existed in v5.

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -1991,43 +1991,40 @@ const runTests = (baseopts: chokidar.ChokidarOptions) => {
     });
   });
 
-  describe('reproduction of bug in vite issue #22672', () => {
-    // https://github.com/vitejs/vite/issues/22672
-    it('should close the fs.watch handle of a deleted watched directory', async () => {
-      if (!isWindows) return true; // separator mismatch only occurs on Windows
+  it('should close the fs.watch handle of a deleted watched directory', async () => {
+    const id = testId.toString();
+    const watchedDir = sp.join(FIXTURES_PATH, id, 'to-delete');
+    await mkdir(watchedDir, { recursive: true });
+    await write(sp.join(watchedDir, 'a.txt'), 'a');
+    await delay();
 
-      const id = testId.toString();
-      const watchedDir = sp.join(FIXTURES_PATH, id, 'to-delete');
-      await mkdir(watchedDir, { recursive: true });
-      await write(sp.join(watchedDir, 'a.txt'), 'a');
-      await delay();
-
-      const watcher = cwatch(watchedDir, options);
-      try {
-        await waitForWatcher(watcher);
-
-        // We assert on the internal `_closers` map because there is no public
-        // signal that distinguishes a closed watcher from a leaked one here
-        const closers = (watcher as unknown as { _closers: Map<string, unknown[]> })._closers;
-        const dirHasCloser = () =>
-          [...closers.keys()].some((key) => sp.resolve(key) === sp.resolve(watchedDir));
-
-        // Sanity check: the watched directory has a registered fs.watch closer.
-        ok(dirHasCloser(), 'expected a closer to be registered for the watched directory');
-
-        const unlinkDirSpy = createSpy<EmitArgs, void>(function unlinkDirSpy() {});
-        watcher.on(EV.UNLINK_DIR, unlinkDirSpy);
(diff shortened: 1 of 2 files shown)


sapphi-red:
The CI failure is probably caused by https://github.com/nodejs/node/issues/63638

43081j:
it is. don't worry about it.

i'll take a look at the change soon 👍

src/index.test.ts
```diff
@@ -1991,6 +1991,45 @@ const runTests = (baseopts: chokidar.ChokidarOptions) => {
     });
   });
 
+  describe('reproduction of bug in vite issue #22672', () => {
```
43081j: i don't think we need this extra describe block or any mention of the vite issue at all.

this is just a piece of functionality that wasn't tested before, the `it` should be enough

src/index.test.ts
```diff
@@ -1991,6 +1991,45 @@ const runTests = (baseopts: chokidar.ChokidarOptions) => {
...
+    // https://github.com/vitejs/vite/issues/22672
+    it('should close the fs.watch handle of a deleted watched directory', async () => {
+      if (!isWindows) return true; // separator mismatch only occurs on Windows
```
43081j: we may as well still run it on other OSes. it works _today_ but who knows if it does in future?

src/index.test.ts
```diff
@@ -1991,6 +1991,45 @@ const runTests = (baseopts: chokidar.ChokidarOptions) => {
...
+      await write(sp.join(watchedDir, 'a.txt'), 'a');
+      await delay();
+
+      const watcher = cwatch(watchedDir, options);
+      try {
+        await waitForWatcher(watcher);
+
+        // We assert on the internal `_closers` map because there is no public
+        // signal that distinguishes a closed watcher from a leaked one here
+        const closers = (watcher as unknown as { _closers: Map<string, unknown[]> })._closers;
```
43081j: to avoid the hackery of exposing internals, we should probably just extend the class in the test:

```ts
class FSWatcherWithClosers extends FSWatcher {
  get closers(): Map<string, unknown[]> {
    return this._closers;
  }
}
const watcher = new FSWatcherWithClosers(options);
watcher.add(paths);
WATCHERS.push(watcher);
```

one day i'd like these internals to be private fields, so i'd rather avoid relying on them in new tests

src/index.test.ts
```diff
@@ -1991,6 +1991,45 @@ const runTests = (baseopts: chokidar.ChokidarOptions) => {
...
+        watcher.on(EV.UNLINK_DIR, unlinkDirSpy);
+
+        await rmr(watchedDir);
+        await waitFor([[unlinkDirSpy, 1, [watchedDir]]]);
+
+        // The closer must have been invoked and removed; otherwise the underlying
+        // fs.watch handle is leaked and keeps firing events on the deleted path.
+        ok(!dirHasCloser(), 'fs.watch handle of the deleted watched directory was leaked');
+      } finally {
+        await watcher.close();
```
43081j: assuming you do my other suggestion, this'll already happen by the top level `afterEach` (it closes all `WATCHERS`)

43081j:
change makes sense and is definitely a bug. good catch!

few comments on the tests but otherwise seems good

src/index.test.ts
sapphi-red: ah, I was following the describe block above. I'll update

src/index.test.ts
sapphi-red: Ah, didn't think of that way!

43081j (approved):

43081j:
tests do pass locally and i've dug through the code myself to check this is the right fix - i'm pretty certain it is.

but i don't have a windows machine right now to prove it for sure, and libuv is a mess.

im going to merge but we need to fix this CI nonsense. @paulmillr i wonder if we should just pin jsbt to use an older Node 24 which doesn't have this problem 🤔

sapphi-red:
FWIW the test failed without the changes on my Windows machine and passed with the changes.

PR #1468, opened 2026-07-06
fix: unixify closer paths rather than normalize

43081j:
`normalize` will retain OS-specific path separators which actually
doesn't fix the problem this change originally intended to fix.

By using the same unixify helper we use when adding paths, we're
ensuring the same normalisation applies when reading/removing.

@sapphi-red FYI this is a miss from your branch I think.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -935,7 +935,7 @@ export class FSWatcher extends EventEmitter<FSWatcherEventMap> {
    * Closes only file-specific watchers
    */
   _closeFile(path: Path): void {
-    const key = sp.normalize(path);
+    const key = normalizePathToUnix(path);
     const closers = this._closers.get(key);
     if (!closers) return;
     closers.forEach((closer) => closer());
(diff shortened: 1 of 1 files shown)


sapphi-red:
I think it normalizes in the other way.

`sp.normalise` will change `/` to `\` on Windows while `normalizePathToUnix` will change `\` to `/` on Windows.

sapphi-red:
Since the key of `_closers` is only read/written by `_closeFile` / `_addPathCloser`, I think this doesn't change the behavior.
That said, it's probably better to have this change so that the path is always unixified.

43081j:
> sp.normalise will change / to \ on Windows while normalizePathToUnix will change \ to / on Windows

this is the way i was describing too. when i said it'll retain OS-specific separators, i meant the resulting string will contain your OS' separators. 

i understand now though, you changed the _setter_ to windows-ify the unixified path rather than just unixifying paths when reading.

so it is right in main, it just changed to storing os-specific paths instead of unix paths

PR #1469, opened 2026-07-07
Add support for fs.watch(recursive: true)

paulmillr:
- one native fs.watch(root, { recursive: true }) per watched tree instead of one watcher per directory
- Enabled on Linux/macOS/Windows when usePolling is off; everything else (and a runtime fs.watch throw) falls back to the classic per-directory mode.

## Inner workings

- handler.ts — _handleDirRecursive creates the native watcher and runs a full-depth readdirp scan to build the DirEntry tree and emit initial events. _handleRecursiveEvent normalizes each raw event by lstat-ing the reported path and diffing against the tree: new dir ⇒ addDir + scan (catches trees moved in with contents), new file ⇒ add, tracked file ⇒ change, ENOENT ⇒ unlink/unlinkDir via the existing _remove. All events flow through _emit, so ignored, cwd, atomic, awaitWriteFinish, and depth keep working.
- Correctness subtleties that surfaced during testing: a create arrives as a rename+change pair, so a seen map (mtime/size + a 50ms post-add window) suppresses the duplicate change; ignored directories need explicit ancestor checks per event (per-dir mode gets that structurally); with followSymlinks: false the link itself is treated as a file so an atomic link replacement emits change (live realpath comparison proved racy); a tracked symlink's ENOENT is re-checked after 100ms before declaring unlink; readdirp is stopped from descending into symlinked dirs, which are delegated to the existing per-directory machinery.

## Benchmark (20k files + 2k dirs)

- Time to ready: 455ms => 240ms
- RSS after ready: 280MB => 96MB
- latency: 0.15ms => 0.06ms
- close(): 42ms => 20ms
- 500-file create burst: 95ms => 500ms (slower)
- kernel inotify watches: same

### larger bench

- New mode ready time: 1.78x faster at 100k, 2.10x faster at 400k.
- New mode RSS: 68.5% lower at 100k, 70.0% lower at 400k.
- New mode change latency: about 3.4x faster median.
- New mode close time: about 2.4x-2.5x faster.
- New mode 500-file add churn is slower here: about 9x-12x slower.
- On Linux, inotify watch count is unchanged between modes.

On Linux, Node emulates recursive watching in JS, so kernel watch counts don't drop — the memory/startup wins come from eliminating chokidar's per-path closer/listener machinery. And the create-burst slowdown is entirely Node's emulation layer: raw fs.watch(recursive) took ~558 ms to deliver the same 500 events, slightly slower than chokidar's wrapper. On macOS (FSEvents) and Windows (ReadDirectoryChangesW) it's one true kernel object, so resource wins should be substantially larger there — worth benchmarking before making it a default anywhere.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.bak
 *.log
 /coverage/
+/.bench-tmp/
 /node_modules/
 /test-fixtures/
 /*.d.ts
--- a/README.md
+++ b/README.md
@@ -132,6 +132,7 @@ chokidar.watch('file', {
   ignoreInitial: false,
   ignorePermissionErrors: false,
   usePolling: false,
+  useRecursiveWatch: false,
   alwaysStat: false,
 });
 ```
@@ -179,6 +180,13 @@ chokidar.watch('file', {
   - `binaryInterval` (default: `300`). Interval of file system
     polling for binary files.
     ([see list of binary extensions](https://github.com/sindresorhus/binary-extensions/blob/master/binary-extensions.json))
+- `useRecursiveWatch` (default: `false`). Whether to use a single native
+  `fs.watch(path, { recursive: true })` instance per watched directory tree,
+  instead of one watcher per directory. Uses ~3x less memory and makes startup
+  ~2x faster on big directory trees. Only effective on Linux, macOS and Windows
+  when `usePolling` is disabled; other platforms fall back to per-directory
+  watchers. Note that on Linux, node.js emulates recursive watching in
+  javascript, so the amount of kernel inotify watches stays the same.
 - `alwaysStat` (default: `false`). If relying upon the
   [`fs.Stats`](https://nodejs.org/api/fs.html#fs_class_fs_stats)
   object that may get passed with `add`, `addDir`, and `change` events, set
--- a/benchmark/bench.mjs
+++ b/benchmark/bench.mjs
@@ -1,0 +1,177 @@
+// Benchmark: chokidar per-directory watchers vs useRecursiveWatch
+// Run `npm run build` first: this imports the compiled index.js.
+// Usage: node benchmark/bench.mjs            (parent: builds tree, spawns one child per mode)
+//        node benchmark/bench.mjs <mode> <dir>  (child)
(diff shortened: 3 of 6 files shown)


43081j:
this is looking good 🎉 

we should fix CI before this can land too. i've opened paulmillr/jsbt#6 for that

paulmillr:
Windows / mac benchmarks are required for decision..

43081j:
I can run them on macos and windows 👍

Will post the results here when I get to it

43081j:
MacOS (macbook M2 air):

- Time to ready: 1106ms -> 205ms
- RSS after ready: 234MB -> 19MB
- latency: 0.06ms -> 51ms (apparently FSEvents has a min 50ms window between events which might explain this)
- close(): 231ms -> 0.4ms
- 500-file create burst: 257ms -> 58ms

paulmillr:
@43081j 1M file test would be really interesting, if possible

43081j:
will do when im back on my macbook 👍 

meanwhile, here's windows (not 1M, just the default benchmark):

- Time to ready: 6120ms → 669ms
- RSS after ready: 263MB → 9MB
- latency (median): 10ms → 2.8ms
- close(): 192ms → 0.4ms
- 500-file create burst: 378ms → 15166ms (timed out? we only got 100-200 of the 500 files)

paulmillr:
we'll need to fix file bursts

43081j:
here's 1M on the same macbook:

- Time to ready: 16663ms
- RSS after ready: 102.3MB
- latency: 50.95ms
- close(): 2.6ms
- 500-file create burst: 59ms

without `recursive` it throws because we have too many watchers open.

so it seems time to ready increases quite a bit as the number of files does

43081j:
@paulmillr did you mean to close this one? 👀 

it'd be good to still figure this out as it would solve a lot of the windows problems people see

paulmillr:
i've rewritten the package completely

https://github.com/paulmillr/chokidar/tree/v6

43081j:
Can you open a draft pr when you get chance? I can help review then 👍

PR #1473, opened 2026-07-14
fix: register watcher before dir scan to avoid missing files created during scan

### LINKED ISSUE #1471: Files created in a newly-added directory between its scan and fs.watch registration are silently missed (still present in v5; same root cause as #1112)
**Describe the bug**

When a new directory appears inside a watched tree, `_handleDir` (handler.js) processes it in this order:

1. `await this._handleRead(...)` — the async readdirp scan of the directory
2. only after the scan completes: `_watchWithNodeFs(dir, ...)` — the `fs.watch` registration

Any file created **after the scan reads the directory entries but before the watch registers** is lost: the scan didn't see it and no watcher existed to observe the event. The file gets no `add` until some unrelated later event in that directory triggers a re-scan.

This is the same race reported in #1112 (v3.5.1). That issue was closed as completed, but the ordering is unchanged in v5.0.0 (and current `master`, src/handler.ts) and the bug still reproduces.

A real-world sequence that hits the window is an atomic write into a fresh folder — `mkdir -p`, stage a temp file, `rename` over the target. The scan can catch only the temp file (whose pending `add` is then dropped by `awaitWriteFinish` when it disappears), and the rename to the final name fires in the unwatched gap: the target file is never reported.

A userland workaround is also harder than it looks: `watcher.add(newDir)` cannot recover the missed file, because `add()` re-runs the scan with `initialAdd = true` and `_handleFile` suppresses the `add` event under `ignoreInitial: true` (`if (!(initialAdd && this.fsw.options.ignoreInitial))`). Recovering requires a manual `readdir` reconciled against `getWatched()`.

- Chokidar version: 5.0.0
- Node version: 24.x
- OS version: macOS 15 and Linux (Debian, kernel 6.x) — both reproduce

**To Reproduce:**

The window is a few milliseconds wide, so per #1112's technique, widen it deterministically by injecting a delay in `_handleDir` between the scan and the watch registration (this doesn't cause the race, it only makes it easy to hit):

```diff
--- a/handler.js
+++ b/handler.js
@@ _handleDir
                 await this._handleRead(dir, initialAdd, wh, target, dir, depth, throttler);
                 if (this.fsw.closed)
                     return;
             }
+            if (!initialAdd) await new Promise((res) => setTimeout(res, 3000));
             closer = this._watchWithNodeFs(dir, (dirPath, stats) => {
```

Then:

```js
import { watch } from "chokidar"
import { mkdtemp, mkdir, writeFile } from "node:fs/promises"
import { join } from "node:path"
import { tmpdir } from "node:os"

const root = await mkdtemp(join(tmpdir(), "chokidar-race-"))
const watcher = watch(root, { ignoreInitial: true })
watcher.on("all", (event, path) => console.log(`EVENT ${event} ${path}`))
await new Promise((ready) => watcher.on("ready", ready))

const dir = join(root, "new-dir")
await mkdir(dir)

// Land the file inside the widened window (scan done, watch not yet registered)
await new Promise((r) => setTimeout(r, 1000))
await writeFile(join(dir, "missed.txt"), "hi")

await new Promise((r) => setTimeout(r, 5000))
console.log("--- 5s later: was missed.txt ever added? ---")

// An unrelated later write triggers a re-scan, which finally surfaces it
await writeFile(join(dir, "second.txt"), "hi")
await new Promise((r) => setTimeout(r, 2000))
await watcher.close()
```

**Expected behavior**

An `add` event for `new-dir/missed.txt` shortly after it is created.

**Actual output**

```
EVENT addDir /tmp/chokidar-race-XXXX/new-dir
--- 5s later: was missed.txt ever added? ---
EVENT add /tmp/chokidar-race-XXXX/new-dir/missed.txt   <- only after second.txt forces a re-scan
EVENT add /tmp/chokidar-race-XXXX/new-dir/second.txt
```

**Possible fix**

Register the `fs.watch` before scanning (the existing tracked-children bookkeeping already dedupes an entry seen by both the scan and an event), or re-run `_handleRead` once immediately after the watch registration to reconcile anything that landed in the gap.

xianjianlf2:
## What

Fixes a race where a file created inside a newly-added directory during that directory's initial scan is silently missed (no `add` event). This is a regression of the same race fixed for #1112 — the ordering had drifted back to scan-then-watch.

## Why

In `src/handler.ts` `_handleDir`, the initial `readdirp` scan (`_handleRead`) ran to completion **before** the fs watcher was attached via `_watchWithNodeFs`. A file created inside the directory during that window was reported by neither side: the scan had already listed the (empty) directory, and the watcher was not attached yet.

## How

Attach the watcher **before** the scan, so an in-scan creation fires the (already-live) watcher, which re-reads the directory.

Deduplication is inherent: `_handleRead`'s existing `previous`/`current` diff only emits `add` for entries not already seen, so files the initial scan already reported never produce a duplicate `add`. A guard releases the freshly-created watcher if the `FSWatcher` is closed mid-scan, to avoid leaking the handle.

## Testing

- Added a regression test in `src/index.test.ts` that deterministically widens the window (wraps `_handleRead` for the new subdir to create a file mid-scan), then asserts exactly one `add` fires — passing under both `fs.watch` (non-polling) and `fs.watchFile` (polling).
- Before: the test times out, the `add` never fires. After: full suite `209 passing`, 0 failing. Verified the test exercises the built code (revert `handler.ts` → red, restore → green).

Closes #1471

### DIFF
--- a/src/handler.ts
+++ b/src/handler.ts
@@ -651,17 +651,36 @@ export class NodeFsHandler {
 
     const oDepth = this.fsw.options.depth;
     if ((oDepth == null || depth <= oDepth) && !this.fsw._symlinkPaths.has(realpath)) {
-      if (!target) {
-        await this._handleRead(dir, initialAdd, wh, target, dir, depth, throttler);
-        if (this.fsw.closed) return;
-      }
-
+      // Register the fs watcher BEFORE running the initial readdirp scan.
+      //
+      // Previously the scan (`_handleRead`) ran to completion and only then was
+      // the watcher attached. That left a window in which a file created inside
+      // `dir` while the scan was in progress was reported by neither side: the
+      // scan had already listed the directory contents (before the file
+      // existed) and the watcher was not attached yet, so the `add` event was
+      // silently lost (#1471 — a regression of the same race fixed in #1112).
+      //
+      // Attaching the watcher first means any creation during the scan fires the
+      // watcher, which re-reads the directory. The `previous`/`current` diff in
+      // `_handleRead` dedupes against entries the initial scan already emitted,
+      // so files reported by the scan do not produce a duplicate `add`.
       closer = this._watchWithNodeFs(dir, (dirPath, stats) => {
         // if current directory is removed, do nothing
         if (stats && stats.mtimeMs === 0) return;
 
         this._handleRead(dirPath, false, wh, target, dir, depth, throttler);
       });
+
+      if (!target) {
+        await this._handleRead(dir, initialAdd, wh, target, dir, depth, throttler);
+        if (this.fsw.closed) {
+          // The watcher was attached before the scan; if we were closed during
+          // the scan its closer will not be registered, so release it here to
+          // avoid leaking the underlying fs_watch handle.
+          if (closer) closer();
+          return;
+        }
+      }
     }
     return closer;
   }
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -559,6 +559,45 @@ const runTests = (baseopts: chokidar.ChokidarOptions) => {
       ok(spy.calls[0][1]); // stats
       ok(rawSpy.called);
     });
+    it('should notice a file created in a new directory during its scan (#1471)', async () => {
+      const testDir = dpath('subdir');
+      const testPath = dpath('subdir/add.txt');
+      const spy = createSpy<EmitArgs, void>(function addSpy() {});
+      watcher.on(EV.ADD, spy);
+      equal(spy.called, false);
+
+      // Deterministically widen the race window from #1471. When the freshly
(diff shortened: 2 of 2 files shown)


xianjianlf2:
Checked the red CI job ("jsbt v0.6.0 / Node v24 @ macos-latest"). The new #1471 regression test passed there; the run failed later in an existing cwd watcher test: "fs.watch (non-polling) > watch options > cwd > should allow separate watchers to have different cwds". The assertion missed the "unlink.txt" event at "index.test.js:1382". This looks unrelated to this PR handledir watcher-order change. I also checked recent main CI and it is not fully clean either, with a separate watcher failure on Windows/Node 24.

PR #1477, opened 2026-08-08
fix: settle directory read when readdirp stream is destroyed

### LINKED ISSUE #1476: A symlink whose target path passes through a regular file (ENOTDIR) silently disables watching of its directory and prevents `ready`
**Describe the bug**

If a watched directory contains a symlink whose *target path passes through a
regular file* (so `realpath()` fails with `ENOTDIR`), chokidar silently stops
watching that directory and the watcher's `ready` event never fires. No
`error` event is emitted, and no events are ever delivered for sibling files
in that directory.

A merely *dangling* symlink (`ENOENT`) behaves fine, which makes this easy to
misdiagnose. Links like this occur in practice when build output changes
shape (e.g. a stale `dist/lib.js/index.js` link after `dist/lib.js` became a
file).

 - Chokidar version: 5.0.0 (readdirp 5.1.1); also reproduced on 4.0.3
 - Node version: 26.7.0
 - OS version: Linux (arm64); also reproduced under Bun 1.3.14

**To Reproduce:**

```js
import { mkdtempSync, writeFileSync, symlinkSync } from "node:fs";
import { join } from "node:path";
import { tmpdir } from "node:os";
import chokidar from "chokidar";

const root = mkdtempSync(join(tmpdir(), "enotdir-"));
writeFileSync(join(root, "regular.txt"), "hi\n");
// Target path passes THROUGH regular.txt → realpath() rejects with ENOTDIR.
// (Replace with a dangling target, e.g. join(root, "missing"), and
// everything works — that's the control case.)
symlinkSync(join(root, "regular.txt", "nope"), join(root, "poison"));

const w = chokidar.watch(root, { followSymlinks: false });
w.on("ready", () => console.log("ready"));
w.on("error", (e) => console.log("error", e.code));
w.on("all", (ev, p) => console.log(ev, p));
setTimeout(() => writeFileSync(join(root, "regular.txt"), "edit\n"), 500);

// Expected: "ready", then add/change for regular.txt (or at minimum an
//           "error" so the caller can react).
// Observed: only the initial addDir for the root. No ready, no error, no
//           add/change for regular.txt — ever.
```

Happy to contribute this as a test case for the test suite if that's useful.

**Expected behavior**

Either treat a failed `realpath()` on a symlink entry the same as the ENOENT
case (skip the entry, keep watching the rest of the directory, fire `ready`),
or surface an `error` event — anything but silent partial coverage with
`ready` withheld.

**Additional context**

The mechanism appears to be in readdirp rather than chokidar itself:
`_getEntryType()` calls `realpath()` on symlink entries and routes failures to
`_onError()`, whose `NORMAL_FLOW_ERRORS` set (`ENOENT`, `EPERM`, `EACCES`,
`ELOOP`, recursive) turns known errnos into `warn` — but anything else,
including `ENOTDIR`, calls `destroy(err)` on the stream (readdirp 5.1.1
`index.js`). The destroyed listing takes the directory's watches and the
pending `ready` accounting with it. A one-line fix may be adding `ENOTDIR` to
the survivable set — a symlink whose target can't resolve isn't a directory to
descend either way — but I'm happy to move this issue to the readdirp repo or
send a PR to either project if you'd prefer.

Found while building [peruse](https://github.com/talkasab/peruse) (a local directory viewer that watches served
directories); we currently work around it by detecting the stalled scan and
covering affected directories with `fs.watch`.

teddytennant:
Fixes #1476.

## The bug

`_handleRead` returns a promise that is resolved only from the stream's `end`
listener (`src/handler.ts`). A readdirp stream that gets destroyed by an error
never emits `end`, so that promise stays pending forever. `_handleDir` awaits it,
so:

- the `_watchWithNodeFs` call after the `await` never runs — the directory ends up
  with no watcher at all,
- the pending ready count is never decremented, so `ready` never fires for the
  whole watcher, not just that directory.

Nothing is reported either, so the caller sees a watcher that is simply stuck.

The reachable trigger from #1476 is a symlink whose target path descends into a
regular file. readdirp calls `realpath()` on symlink entries and only tolerates a
fixed set of errnos (`ENOENT`, `EPERM`, `EACCES`, `ELOOP`, recursion); anything
else calls `destroy(err)`. `ENOTDIR` is not in that set, so one such link kills
the listing of its whole directory. A merely dangling link (`ENOENT`) is fine,
which makes this easy to misdiagnose.

Reproducer (from the issue, verified on `main`):

```js
import { mkdtempSync, writeFileSync, symlinkSync } from 'node:fs';
import { join } from 'node:path';
import { tmpdir } from 'node:os';
import chokidar from 'chokidar';

const root = mkdtempSync(join(tmpdir(), 'enotdir-'));
writeFileSync(join(root, 'regular.txt'), 'hi\n');
symlinkSync(join(root, 'regular.txt', 'nope'), join(root, 'poison'));

const w = chokidar.watch(root, { followSymlinks: false });
w.on('ready', () => console.log('READY'));
w.on('error', (e) => console.log('ERROR', e.code));
w.on('all', (ev, p) => console.log('EVENT', ev, p));
setTimeout(() => w.close(), 3000);
```

Before: only `EVENT addDir <root>`, then nothing — no `READY`, ever.
After: `EVENT addDir <root>` followed by `READY`.

## The fix

Settle the promise on `close` as well as `end`. `close` follows both a normal
`end` (readdirp sets `autoDestroy: true`) and a `destroy(err)`, so the listing
promise now resolves however the stream terminates. `resolve` is idempotent, so
the normal path is unchanged — the existing `end` listener still does the
snapshot-diff / `_remove` / throttle-retry work first, and that work is
deliberately *not* run for a destroyed stream, since a truncated listing would
make the diff unlink files that still exist.

This also covers the second non-settling path in the same block: the `end`
listener returns early without resolving when `this.fsw.closed` is already true.

## Why here

The issue also points at readdirp, and that is fair — adding `ENOTDIR` to
readdirp's `NORMAL_FLOW_ERRORS` would let the rest of that directory be listed,
which this patch does not do on its own. But the two are independent. Whatever
errno readdirp decides to destroy a stream over, chokidar should not respond by
hanging forever with no `ready` and no error, and that is a chokidar-side defect
in `_handleRead`. Happy to follow up on the readdirp side separately if you want
the entries after the bad link to be picked up too.

Nothing else was changed: `_handleError` still deliberately swallows `ENOTDIR`
(and `ENOENT`) rather than emitting `error`, so this does not add a new error
event to anyone's watcher.

## Test

`should emit ready event when a symlink target path passes through a file`, next
to the existing broken-symlink ready test in `watch symlinks`. It runs under both
the polling and non-polling variants.

Fail-before (fix reverted, test only):

```
chokidar
└─fs.watchFile (polling)
  ├─watch symlinks
  │ └─should emit ready event when a symlink target path passes through a file: ☓
Error: timeout
    at Timeout._onTimeout (file:///.../index.test.js:70:20)
```

Pass-after:

```
chokidar
└─fs.watchFile (polling)
  ├─watch symlinks
  │ └─should emit ready event when a symlink target path passes through a file: ✓

1 tests passed in 1 sec
```

Full suite, Linux / Node 22: `207 tests passed` on `main`, `209 tests passed`
with this branch (the new test counts once per variant). Ran five times on the
branch, clean each time. `prettier --check src` and `tsc` are clean.

Note for context: the suite is flaky on my machine *without* this patch — two
of four `main` runs failed, on `should ignore unwatched siblings` and
`should close the fs.watch handle of a deleted watched directory` respectively.
That is pre-existing and unrelated; I did not touch it.

### DIFF
--- a/src/handler.ts
+++ b/src/handler.ts
@@ -588,6 +588,13 @@ export class NodeFsHandler {
 
     return new Promise((resolve, reject) => {
       if (!stream) return reject();
+      // A stream destroyed by an error never emits `end`, and neither does one
+      // abandoned after the watcher was closed. `close` always follows, so settle
+      // there too: leaving this pending stalls _handleDir and the ready counter.
+      stream.once(STR_CLOSE, () => {
+        stream = undefined;
+        resolve(undefined);
+      });
       stream.once(STR_END, () => {
         if (this.fsw.closed) {
           stream = undefined;
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -1123,6 +1123,16 @@ const runTests = (baseopts: chokidar.ChokidarOptions) => {
       await waitForWatcher(watcher);
       equal(readySpy.callCount, 1);
     });
+    it('should emit ready event when a symlink target path passes through a file', async () => {
+      options.followSymlinks = false;
+      // Resolving this link fails with ENOTDIR rather than the ENOENT of a
+      // dangling one, because its target path descends into a regular file.
+      await symlink(dpath('subdir/add.txt/nope'), dpath('subdir/notdir'));
+      const readySpy = createSpy(function readySpy() {});
+      const watcher = cwatch(dpath('subdir'), options).on(EV.READY, readySpy);
+      await waitForWatcher(watcher);
+      equal(readySpy.callCount, 1);
+    });
   });
   describe('watch arrays of paths/globs', () => {
     it('should watch all paths in an array', async () => {


teddytennant:
The 9 red jobs here are not 9 failures. The matrix is fail-fast, so 8 of them are cancellations — every one of them is annotated `The strategy configuration was canceled because "test-ts.node._26_windows-latest" failed`.

The one job that actually failed is Windows, and it dies on a libuv assertion rather than a test:

```
183 tests started...
chokidar
├─should expose public API methods: ✓
├─fs.watch (non-polling)
│ ├─watch a directory
│ │ ├─should produce an instance of chokidar.FSWatcher: ✓
│ │ ├─should expose public API methods: ✓
│ │ ├─should emit `add` event when file was added: ☆
Assertion failed: !_wcsnicmp(filename, dir, dirlen), file src\win\fs-event.c, line 72
##[error]Process completed with exit code 1.
```

https://github.com/paulmillr/chokidar/actions/runs/31275748945/job/93148878722

`main` at this PR's merge base (0bc7bed) fails identically — same assertion, same test, third test of the run, whole matrix cancelled behind it:

https://github.com/paulmillr/chokidar/actions/runs/28794759888/job/85382171254

That is three tests in, in `fs.watch (non-polling) > watch a directory`, long before anything this branch touches, and the abort comes from libuv's own `src/win/fs-event.c`.

The other 8 jobs never report a failing test — grepping all 8 logs for `☓` returns zero hits. Each was mid-run and green when it was killed. Node 22 @ ubuntu-26.04-arm's last line before `The operation was canceled` is a passing test at 19:59:15, sixteen seconds after the Windows job crashed at 19:58:59; Node 22 @ macos-latest likewise. So the cancellations are not a hang or a timeout.

Locally on Linux / Node 22.22, same steps as CI (`npm install`, `npm run build`, `npm test` with `MSHOULD_FAST=1`):

- merge base 0bc7bed: `207 tests passed`
- this branch: `209 tests passed`, four consecutive runs

`prettier --check src` and `tsc` are clean.

Happy to rebase if you land a fix for the Windows crash and want a green run on this.

43081j:
assuming you're using an agent for these responses. if you are, let's not, and converse with you directly if possible 👍 

or at the very least let us know this is the case so we know to take extra care reviewing.

anyhow, the true fix seems to be to add ENOTDIR to readdirp here, as you noticed:
https://github.com/paulmillr/readdirp/blob/5dba3694a1de891ea6edb0301240326f8f8c81f1/index.ts#L81

the change here is more of a bandaid, really. though i think it is still fine to land.

teddytennant:
Makes sense, readdirp is the right layer for it. Happy to send the ENOTDIR patch there if you want it, and this one can either stay as a guard on the chokidar side or close once that lands, whichever you prefer.

PR #1479, opened 2026-08-14
chore: bump jsbt workflows

43081j:
Hopefully this fixes the libuv nonsense :eyes:


43081j:
bun is still a bit of a mess in regards to FS implementation. remember we turned off the tests because of that.

i'm tempted to set `node-only` here. what do you think? our build script already builds the typescript too, so we are testing against the built test file.

PR #1480, opened 2026-08-23
V6

paulmillr:


paulmillr:
The goal is to release it tomorrow.

43081j:
I've been away all week so haven't had much chance to read this properly yet. I'd really like to finish reading through it as it's a huge change. 

I'm still away but I'll try have a look today

src/index.ts
```diff
@@ -318,106 +128,152 @@ export interface FSWatcherEventMap {
...
+  private events: EventPolicy;
+
+  get closed(): boolean {
+    return this.lifecycle !== undefined && this.lifecycle.state !== 'OPEN';
+  }
+
+  private ignoredPaths: Set<Matcher>;
+  private streams: Set<ReaddirpStream>;
+
+  private pendingAdds: Map<string, symbol>;
```
43081j, line 144: nit: we should probably just use native private fields instead, i.e. `#pendingAdds: Map<string, symbol>;`

src/index.ts
```diff
@@ -318,106 +128,152 @@ export interface FSWatcherEventMap {
...
-      atomic: true, // NOTE: overwritten later (depends on usePolling)
-      ..._opts,
+      atomic: _opts.atomic ?? true,
       // Change format
-      ignored: _opts.ignored ? arrify(_opts.ignored) : arrify([]),
+      ignored: Object.freeze(_opts.ignored ? arrify(_opts.ignored).map(cloneOwnedMatcher) : []),
       awaitWriteFinish:
-        awf === true ? DEF_AWF : typeof awf === 'object' ? { ...DEF_AWF, ...awf } : false,
+        awf === true
+          ? Object.freeze({ ...DEF_AWF })
```
43081j, line 201: may be worth just freezing DEF_AWF up front and returning it as-is, so we only allocate and freeze when you provide an `awf` object yourself

src/index.ts
```diff
@@ -318,106 +128,152 @@ export interface FSWatcherEventMap {
...
+    opts.interval = opts.pollingInterval;
+    opts.binaryInterval = opts.pollingBinaryInterval;
+    opts.usePolling = opts.backend === 'polling';
+    opts.backendCapabilities = selectBackend(opts);
+    opts.backendStrategy = opts.backendCapabilities.kind;
+    // Editor atomic write normalization is enabled by default only with fs.watch.
+    // Inspect the raw option so the merged default cannot hide an implicit choice.
+    if (_opts.atomic === undefined) opts.atomic = !opts.usePolling;
+    validateOptions(opts);
+    this.emitRaw = (...args) => {
```
43081j, line 235: given this is private anyway, can we just make it a method? 👀

src/index.ts
```diff
@@ -318,106 +128,152 @@ export interface FSWatcherEventMap {
...
+      isPathGenerationActive: (path, generation) => this.isPathGenerationActive(path, generation),
+      isReady: () => this.readyEmitted,
+      remove: (directory, item) => this.removePath(directory, item),
+      handleError: (error) => this.handleError(error),
+      publish: (event, args) => this.emitWithAll(event, args as EmitArgs),
+    });
+    void this.emitRaw;
+    void this.createHelper;
+    void this.addPathCloser;
+    void this.createScanStream;
```
43081j, line 266: are these still needed? assuming they were here to silence unused-properties warnings during dev

src/index.ts
43081j: aha i see they're not actually private... the observation code for example calls some of these. that's awkward.

maybe lets just be honest and make the ones we reference outside `public`? they are being used as public fields right now anyway.

then we won't need to forcefully cast it when passing it into the observation engine too

src/index.ts
```diff
@@ -590,383 +484,260 @@ export class FSWatcher extends EventEmitter<FSWatcherEventMap> {
...
-    this._getWatchedDir(dir).remove(sp.basename(path));
+  private closePath(path: Path, recursive = false): void {
+    const logicalKey = logicalPathKey(path);
+    const contains = (candidate: string): boolean => {
+      if (candidate === logicalKey) return true;
+      if (!recursive) return false;
+      return isStrictlyInside(logicalKey, candidate);
+    };
+
+    [...this.lifecycle.closers.keys()].filter(contains).forEach((key) => this.closeFile(key));
```
43081j, line 674: for these few iterables, you can probably save some memory/gc by not allocating these arrays, i.e. just do them in a `for...of` and the `filter` becomes a regular `if`.

not a blocker though

src/index.ts
```diff
@@ -590,383 +484,260 @@ export class FSWatcher extends EventEmitter<FSWatcherEventMap> {
...
-  /**
-   * Provides directory tracking objects
-   * @param directory path of the directory
-   */
-  _getWatchedDir(directory: string): DirEntry {
-    const dir = sp.resolve(directory);
-    if (!this._watched.has(dir)) this._watched.set(dir, new DirEntry(dir, this._boundRemove));
-    return this._watched.get(dir)!;
+  private removeTreeItem(directory: string, item: string): void {
+    const entry = this.tree.getDirectory(directory);
```
43081j, line 576: this inserts a directory entry for the directory of the file being removed.

i.e. `/foo/bar` being removed would add a directory entry for `/foo`.

should we maybe check if there's an entry but not upsert one?

src/index.ts
```diff
@@ -590,383 +484,260 @@ export class FSWatcher extends EventEmitter<FSWatcherEventMap> {
...
-      const ignoredPaths = [...this._ignoredPaths];
-      const list: Matcher[] = [...ignoredPaths.map(normalizeIgnored(cwd)), ...ignored];
-      this._userIgnored = anymatch(list, undefined);
+      const ignored = (ign || []).map((matcher) => normalizeMatcher(matcher, cwd));
+      const direct = compileMatchers(ignored);
+      const pathMatchers = ignored.filter(
+        (matcher) => typeof matcher === 'string' || isMatcherObject(matcher)
+      );
+      const canonical = compileMatchers(pathMatchers);
+      const pathAliases = new Map<string, string>();
```
43081j, line 524: this seems like it will grow with the number of paths we check are ignored.

maybe we can simplify this a bunch by just memoizing real paths. like:

```ts
const realPaths = new Map<string, string | undefined>();
const getRealDir = (p: string): string | undefined => {
  // if its in realPaths, return it
  // otherwise, do the try/catch and cache the result in realPaths
};

// elsewhere...

const realDir = getRealDir(sp.dirname(absoluteCandidate));
if (realDir === undefined) return false;
const realPath = sp.join(realDir, sp.basename(absoluteCandidate));
return realPath !== absoluteCandidate && canonical(realPath, candidateStats);
```

PR #1481, opened 2026-09-04
Route a rejection from add() into the error event instead of an unhandled rejection

### LINKED ISSUE #1378: Errors within `add` bubble as unhandled rejections
**Describe the bug**

 - Chokidar version [e.g. 3.2.1 or commit hash] 3.6.0
 - Node version [e.g. 12.11.0, ensure you are using the latest node.js] 20.9.0
 - OS version: [e.g. Ubuntu 19.04 or MacOS 10.15 or Windows 10]

**To Reproduce:**

We have a number of these stack traces where an error occurs within the async function [on this line](https://github.com/paulmillr/chokidar/blob/97894d36657c8e4571bd1a07fb6e2a3f50f7019d/src/index.ts#L466):
```
Error: ENOSPC: System limit for number of file watchers reached, watch '/home/runner/Foodsource5/add_test_data.py'
  File "node:internal/fs/watchers", line 247, col 19, in FSWatcher.<computed>
  File "node:fs", line 2392, col 36, in Object.watch
  File "../../../node_modules/.pnpm/chokidar@3.6.0/node_modules/chokidar/lib/nodefs-handler.js", line 119, col 15, in createFsWatchInstance
    return fs.watch(path, options, handleEvent);
  File "../../../node_modules/.pnpm/chokidar@3.6.0/node_modules/chokidar/lib/nodefs-handler.js", line 166, col 15, in setFsWatchListener
    watcher = createFsWatchInstance(
  File "../../../node_modules/.pnpm/chokidar@3.6.0/node_modules/chokidar/lib/nodefs-handler.js", line 331, col 14, in NodeFsHandler._watchWithNodeFs
    closer = setFsWatchListener(path, absolutePath, options, {
  File "../../../node_modules/.pnpm/chokidar@3.6.0/node_modules/chokidar/lib/nodefs-handler.js", line 395, col 23, in NodeFsHandler._handleFile
    const closer = this._watchWithNodeFs(file, listener);
  File "../../../node_modules/.pnpm/chokidar@3.6.0/node_modules/chokidar/lib/nodefs-handler.js", line 637, col 21, in NodeFsHandler._addToNodeFs
    closer = this._handleFile(wh.watchPath, stats, initialAdd);
  File "../../../node_modules/.pnpm/chokidar@3.6.0/node_modules/chokidar/index.js", line 451, col 21, in <anonymous>
    const res = await this._nodeFsHandler._addToNodeFs(path, !_internal, 0, 0, _origAdd);
  File "index 0", in Promise.all
```

It bubbles, but since `add` is sync, and there is [no `.catch` on this line](https://github.com/paulmillr/chokidar/blob/97894d36657c8e4571bd1a07fb6e2a3f50f7019d/src/index.ts#L481), its thrown to the global unhandled rejection handler. Would it be possible to update the API to allow callsites to handle this error more locally? Either make `add` async, add an error callback parameter, etc.?

I checked v4.0 and it appears the same code is still in place. 

Thank you!

Steps to reproduce the behavior. Include filename and chokidar config.

Ideally prove a problem by isolating and making it reproducible with a very short sample program, which you could paste here:

```
const fs = require('fs');
// One-liner for files and directories starting with 'test'
chokidar.watch('test*', {}).on('all', (event, path) => {
  console.log(event, path);
});
fs.writeFileSync('test.txt', 'testing 1');
// In a comment describe expected output versus observed problem
```

Most valuable could be one or more test cases for [test.js](https://github.com/paulmillr/chokidar/blob/master/test.js) to demonstrate the problem.

**Expected behavior**
A clear and concise description of what you expect to happen.

**Additional context**
Optionally nice to know what project you are working on.

afonsojanu:
Fixes #1378.

`add()` kicks off `Promise.all(...)` internally but returns synchronously, so a rejection from `_addToNodeFs` (an ENOSPC when the OS runs out of file watchers, per the issue's stack trace) had nowhere to go and became a global unhandled rejection.

Since making `add()` async would be a breaking change, I went with what @43081j suggested on the thread: route it through the existing `_handleError()` path instead, same mechanism every other internal failure in this class already uses. Benign errors (ENOENT, ENOTDIR, permission errors when configured) still get filtered out there, so this only changes behavior for genuine failures.

Added a regression test that patches `_addToNodeFs` to reject and confirms an `error` event fires instead of a process-level unhandled rejection. Full suite passes locally (209 tests, both the plain and polling variants).

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -1991,6 +1991,43 @@ const runTests = (baseopts: chokidar.ChokidarOptions) => {
     });
   });
 
+  describe('reproduction of bug in issue #1378', () => {
+    it('should emit an error event instead of an unhandled rejection when add() fails internally', async () => {
+      // Watch nothing initially so the only _addToNodeFs call under test is
+      // the explicit watcher.add() below, not an unrelated directory scan.
+      const watcher = cwatch([], { persistent: true });
+      const unhandled: unknown[] = [];
+      const onUnhandled = (reason: unknown) => unhandled.push(reason);
+      process.on('unhandledRejection', onUnhandled);
+      try {
+        const caught = new Promise<Error>((resolve, reject) => {
+          const timeout = setTimeout(
+            () => reject(new Error('timed out waiting for an error event')),
+            2000
+          );
+          watcher.once('error', (err: unknown) => {
+            clearTimeout(timeout);
+            resolve(err as Error);
+          });
+        });
+        // Simulate the ENOSPC-style failure reported in the issue: something
+        // inside _addToNodeFs rejects after add() has already returned
+        // synchronously, so the caller has no way to .catch() it directly.
+        (watcher as any)._nodeFsHandler._addToNodeFs = async () => {
+          throw new Error('simulated ENOSPC from _addToNodeFs');
+        };
+        watcher.add(sp.join(currentDir, 'somefile-for-1378'));
+        const error = await caught;
+        ok(error.message.includes('simulated ENOSPC from _addToNodeFs'));
+        await delay(50);
+        deepEqual(unhandled, []);
+      } finally {
+        process.off('unhandledRejection', onUnhandled);
+        watcher.close();
+      }
+    });
+  });
+
   it('should close the fs.watch handle of a deleted watched directory', async () => {
     const id = testId.toString();
     const watchedDir = sp.join(FIXTURES_PATH, id, 'to-delete');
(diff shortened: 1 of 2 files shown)


afonsojanu:
Heads up, the failing checks here aren't related to this change: same jsbt matrix has been red on main since at least July, and the one failure I looked at closely (Windows) is an assertion inside libuv's own fs-event.c, nothing to do with add()'s promise handling. Leaving this as is rather than trying to fix an unrelated, pre-existing CI issue in this PR.

43081j (approved):
approving but holding off merging yet in case the new major makes this redundant.

PR #1483, opened 2026-09-07
fix: close a dir watcher when close() lands during its add

### LINKED ISSUE #1482: close() during an in-flight add drops the closer: the fs.watch handle is orphaned and the process never exits
## The race

`_addToNodeFs()` registers the closer returned by `_handleDir()` with `_addPathCloser()` **only after** re-checking `this.fsw.closed`:

```ts
closer = await this._handleDir(...);
if (this.fsw.closed) return;   // ← bare return: closer discarded
...
if (closer) this.fsw._addPathCloser(path, closer);
```

When `close()` lands inside that await gap, it sweeps a still-empty `_closers` list, the resumed `_addToNodeFs` discards the closer, and the `fs.watch` handle created moments earlier is orphaned — nothing will ever close it. Pending `FSReqCallback`s from the in-flight scan keep running too. The process stays alive indefinitely.

## Observed in the wild

apache/maka CI (vite 8.2.2, which bundles a pnpm-patched chokidar 3.6.0): a dev server restarted by a manifest edit kept scanning in the background; `server.close()` raced it; the orphaned `fs.watch` on the temp repository root kept a `node --test` child process alive for 13.5 minutes after **all tests had passed and reported**, until the CI workspace hit its 900s ceiling. Full fs-level attribution (`FSEventWrap` + pending `FSReqCallback` via `process.getActiveResourcesInfo()`, watcher creation stack) and a ~6% natural-occurrence repro under load: https://github.com/apache/maka/issues/4940

## Affects

- **main** (v5): verified on `src/handler.ts` — both `await this._handleDir(...)` call sites (directory branch and symlink-parent branch) are followed by a bare `if (this.fsw.closed) return;` before registration.
- **3.6.0**: same structure (`lib/nodefs-handler.js`); vendored by vite via `patches/chokidar@3.6.0.patch`, which does not cover this.
- An independent forced-interleave reproduction confirms the drop point: `closed = true`, closers list empty, the native `fs.watch` still open; invoking the lost closer by hand closes it.

## Proposed fix (PR follows)

At both drop points, dispose the just-created watcher instead of returning past it:

```ts
if (this.fsw.closed) {
  if (closer) closer();
  return;
}
```

orangeCatDeveloper:
Fixes #1482

## Summary

`_addToNodeFs()` registers the closer returned by `_handleDir()` with `_addPathCloser()` only after re-checking `this.fsw.closed`. When `close()` lands inside that await gap, it sweeps a still-empty `_closers` list, the resumed `_addToNodeFs` discards the closer, and the `fs.watch` handle created moments earlier is orphaned — nothing ever closes it and the process stays alive indefinitely.

This adds one disposal at each of the two drop points (directory branch, symlink-parent branch): invoke the closer instead of returning past it. The `_handleFile` path is synchronous and has no gap, so it needs nothing.

Details, real-world impact (a vite dev-server suite hung on CI with every test passing), and an independent forced-interleave reproduction of the exact drop point: #1482, https://github.com/apache/maka/issues/4940

## Verification

- New regression test `close during scan > closes a dir handle whose closer close() can no longer register`: it forces `close()` to land between the directory `fs.watch` creation and the closer registration, then asserts no `FSEventWrap` survives.
- With the fix: **209 tests passed** (node 24, Linux docker; same on macOS node 26).
- Without the fix (both sites reverted, verified 2 → 0): the new test **fails** on Linux — the orphaned handle is detected.
- `npm run lint` and `npm run build` clean.

Note for downstream: vite vendors chokidar 3.6.0 via `patches/chokidar@3.6.0.patch` and 3.6.0 is EOL, so their patch needs the same disposal for bundled-3.6.0 users until they can bump.

Generated-by: GLM-5.3-Flash (pi)

### DIFF
--- a/src/handler.ts
+++ b/src/handler.ts
@@ -718,7 +718,10 @@ export class NodeFsHandler {
           wh,
           targetPath
         );
-        if (this.fsw.closed) return;
+        if (this.fsw.closed) {
+          if (closer) closer();
+          return;
+        }
         // preserve this symlink's target path
         if (absPath !== targetPath && targetPath !== undefined) {
           this.fsw._symlinkPaths.set(absPath, targetPath);
@@ -730,7 +733,10 @@ export class NodeFsHandler {
         this.fsw._getWatchedDir(parent).add(wh.watchPath);
         this.fsw._emit(EV.ADD, wh.watchPath, stats);
         closer = await this._handleDir(parent, stats, initialAdd, depth, path, wh, targetPath);
-        if (this.fsw.closed) return;
+        if (this.fsw.closed) {
+          if (closer) closer();
+          return;
+        }
 
         // preserve this symlink's target path
         if (targetPath !== undefined) {
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -892,6 +892,26 @@ const runTests = (baseopts: chokidar.ChokidarOptions) => {
       ok(calledWith(spy, [EV.ADD, testPath]));
     });
   });
+  describe('close during scan', () => {
+    it('closes a dir handle whose closer close() can no longer register', async () => {
+      const testDir = dpath('race');
+      await mkdir(testDir);
+      const watcher = new chokidar.FSWatcher(options);
+      const w: any = watcher._nodeFsHandler;
+      const orig = w._watchWithNodeFs.bind(w);
+      const leaks = () =>
+        (process as any).getActiveResourcesInfo().filter((r: string) => r === 'FSEventWrap').length;
+      const baseline = leaks();
+      w._watchWithNodeFs = (path: string, listener: unknown) => {
+        const closer = orig(path, listener);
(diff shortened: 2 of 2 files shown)


orangeCatDeveloper:
Heads-up for review: the red CI here is not from this change. Probe: [pristine `main` + an empty commit](https://github.com/paulmillr/chokidar/pull/1484) fails identically right now — all 9 matrix jobs, with `windows-latest` crashing on the pre-existing libuv assertion `Assertion failed: !_wcsnicmp(filename, dir, dirlen), file src\win\fs-event.c, line 72` at the same test (`watch a directory > should emit `add` event`, before any test of this PR runs), and the other platforms cancelled by fail-fast. The same windows assertion already failed the unrelated #1481 on Sep 4; main was last green Aug 23, so this looks like runner-image/node-24 drift. Once the environment settles, a re-run should be meaningful.

PR #1487, opened 2026-09-15
Drop the removed useFsEvents option from the readme

hxperl:
chokidar 5 has no `useFsEvents` option — the fsevents backend is gone — but the readme still documents it twice, including as what `atomic`'s default depends on. Observed on 0bc7bed:

```
chokidar.watch('.').options.atomic                     -> true
chokidar.watch('.', {usePolling: true}).options.atomic -> true
```

So `atomic` is plain `true`; `usePolling` does not change it. `npm test`: 207 passed.

If you would rather fix that in code than document it, say so and I will close this.

CI is red before this branch: Node 24 / Windows dies on a libuv assertion (`src\win\fs-event.c:72`), same on `main` at 0bc7bed. A readme edit cannot reach it.


43081j:
technically true but im not a fan of agentic spam.

60+ PRs opened in a very short time by your agent. feel free to write a PR description yourself some day and i'll be happy to merge it. otherwise, this smells of spam/rep farming.

